PR #6, opened 2018-04-19
Random eviction support

ms705:
Moves to `rahashmap` for the underlying storage and adds support for random key eviction.

Maybe merge in a branch to insulate consumers against changes that break `rahashmap` (which seem relatively common).


jonhoo:
I wonder whether the thing to do here would be to add this under a feature flag with a hope that the flag can be removed when something like https://internals.rust-lang.org/t/pre-rfc-abandonning-morals-in-the-name-of-performance-the-raw-entry-api/7043 comes to pass.

PR #7, opened 2019-01-16
Add optional features that can potentially improve performance

novacrazy:
This started out as an experiment to see if I could make `evmap` work better with the libraries I'm already using in my projects, but there really isn't a reason not to provide them as features.

The big ones:

[`hashbrown`](https://github.com/Amanieu/hashbrown): Fast and space efficient hash table replacement.
~~[`parking_lot`](https://github.com/Amanieu/parking_lot): Fast and space efficient synchronization primitives.~~

These two are obvious, as they provide generally better and potentially safer functionality (`parking_lot` doesn't poison threads).

Additionally:

[`smallvec`](https://github.com/servo/rust-smallvec): Mixed stack/inline/heap allocated vector.

Although the multi-value map is useful for multiple insertions, if a map never uses that and only ever replaces a single value with `update`, we can cheat by effectively using `[T; 1]` instead, which saves us an extra level of indirection. `smallvec` does grow and allocate if necessary.

Combined, this seems to net about a 50% performance increase on my system, but I need to do a clean run of benchmarks before this is merged. It's certainly better. Plus `hashbrown` is supposed to be more space efficient. 

read-throughput

I switched to log10 because the others were getting lost. Also, I have a 16C/32T CPU, so I bumped up the reader counts to compensate. Although for new benchmarks I may try to change the CPU affinity.

Write speeds are about the same.

What are your thoughts?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,6 +25,6 @@ hashbrown = { version = "0.1.8", optional = true }
 smallvec = { version = "0.6.7", optional = true }
 
 [features]
-default = []
+default = ["hashbrown", "smallvec"]
 nightly_hashbrown = ["hashbrown/nightly"]
 nightly_smallvec = ["smallvec/union"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,11 +22,9 @@ maintenance = { status = "passively-maintained" }
 
 [dependencies]
 hashbrown = { version = "0.1.8", optional = true }
-parking_lot = { version = "0.7.1", optional = true }
 smallvec = { version = "0.6.7", optional = true }
 
 [features]
 default = []
 nightly_hashbrown = ["hashbrown/nightly"]
-nightly_parking_lot = ["parking_lot/nightly"]
 nightly_smallvec = ["smallvec/union"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,3 +21,12 @@ travis-ci = { repository = "jonhoo/rust-evmap" }
 maintenance = { status = "passively-maintained" }
 
 [dependencies]
+hashbrown = { version = "0.1.8", optional = true }
+parking_lot = { version = "0.7.1", optional = true }
+smallvec = { version = "0.6.7", optional = true }
+
+[features]
+default = []
+nightly_hashbrown = ["hashbrown/nightly"]
+nightly_parking_lot = ["parking_lot/nightly"]
+nightly_smallvec = ["smallvec/union"]
--- a/benchmark/Cargo.toml
+++ b/benchmark/Cargo.toml
@@ -4,7 +4,7 @@ version = "0.1.0"
 authors = ["Jon Gjengset <jon@thesquareplanet.com>"]
 
 [dependencies]
-evmap = { path = "../", features = ["hashbrown", "smallvec"] }
+evmap = { path = "../" }
 chashmap = "2.1.0"
 clap = "2.20.3"
 zipf = "0.2.0"
--- a/benchmark/Cargo.toml
+++ b/benchmark/Cargo.toml
@@ -4,7 +4,7 @@ version = "0.1.0"
 authors = ["Jon Gjengset <jon@thesquareplanet.com>"]
 
 [dependencies]
-evmap = { path = "../", features = ["hashbrown", "parking_lot", "smallvec"] }
+evmap = { path = "../", features = ["hashbrown", "smallvec"] }
 chashmap = "2.1.0"
 clap = "2.20.3"
 zipf = "0.2.0"
@@ -18,4 +18,4 @@ codegen-units = 1
 debug = false
 
 [features]
-nightly = ["evmap/nightly_hashbrown", "evmap/nightly_parking_lot", "evmap/nightly_smallvec"]
+nightly = ["evmap/nightly_hashbrown", "evmap/nightly_smallvec"]
--- a/benchmark/Cargo.toml
+++ b/benchmark/Cargo.toml
@@ -11,9 +11,6 @@ zipf = "0.2.0"
 rand = "0.3"
 parking_lot = "0.7.1"
 
-[target.'cgf(not(target_env = "msvc"))'.dependencies]
-jemallocator = {version = "0.1.9", optional = true }
-
 [profile.release]
 lto = true
 opt-level = 3
@@ -21,4 +18,4 @@ codegen-units = 1
 debug = false
 
 [features]
-nightly = ["evmap/nightly_hashbrown", "evmap/nightly_parking_lot", "evmap/nightly_smallvec", "jemallocator"]
+nightly = ["evmap/nightly_hashbrown", "evmap/nightly_parking_lot", "evmap/nightly_smallvec"]
--- a/benchmark/Cargo.toml
+++ b/benchmark/Cargo.toml
@@ -4,11 +4,21 @@ version = "0.1.0"
 authors = ["Jon Gjengset <jon@thesquareplanet.com>"]
 
 [dependencies]
-evmap = { path = "../" }
+evmap = { path = "../", features = ["hashbrown", "parking_lot", "smallvec"] }
 chashmap = "2.1.0"
 clap = "2.20.3"
 zipf = "0.2.0"
 rand = "0.3"
+parking_lot = "0.7.1"
+
+[target.'cgf(not(target_env = "msvc"))'.dependencies]
+jemallocator = {version = "0.1.9", optional = true }
 
 [profile.release]
-debug = true
+lto = true
+opt-level = 3
+codegen-units = 1
+debug = false
+
+[features]
+nightly = ["evmap/nightly_hashbrown", "evmap/nightly_parking_lot", "evmap/nightly_smallvec", "jemallocator"]
--- a/benchmark/src/main.rs
+++ b/benchmark/src/main.rs
@@ -6,13 +6,6 @@ extern crate rand;
 extern crate zipf;
 extern crate parking_lot;
 
-#[cfg(all(not(target_env = "msvc"), feature = "nightly"))]
-extern crate jemallocator;
-
-#[cfg(all(not(target_env = "msvc"), feature = "nightly"))]
-#[global_allocator]
-static ALLOC: jemallocator::Jemalloc = jemallocator::Jemalloc;
-
 use chashmap::CHashMap;
 use std::collections::HashMap;
 use clap::{App, Arg};
--- a/benchmark/src/main.rs
+++ b/benchmark/src/main.rs
@@ -4,6 +4,14 @@ extern crate clap;
 extern crate evmap;
 extern crate rand;
 extern crate zipf;
+extern crate parking_lot;
+
+#[cfg(all(not(target_env = "msvc"), feature = "nightly"))]
+extern crate jemallocator;
+
+#[cfg(all(not(target_env = "msvc"), feature = "nightly"))]
+#[global_allocator]
+static ALLOC: jemallocator::Jemalloc = jemallocator::Jemalloc;
 
 use chashmap::CHashMap;
 use std::collections::HashMap;
@@ -135,7 +143,7 @@ fn main() {
         let (r, w) = evmap::Options::default()
             .with_capacity(5_000_000)
             .construct();
-        let w = sync::Arc::new(sync::Mutex::new((w, 0, refresh)));
+        let w = sync::Arc::new(parking_lot::Mutex::new((w, 0, refresh)));
         let start = time::Instant::now();
         let end = start + dur;
         join.extend((0..readers).into_iter().map(|_| {
@@ -218,7 +226,7 @@ impl Backend for sync::Arc<sync::RwLock<HashMap<u64, u64>>> {
 
 enum EvHandle {
     Read(evmap::ReadHandle<u64, u64>),
-    Write(sync::Arc<sync::Mutex<(evmap::WriteHandle<u64, u64>, usize, usize)>>),
+    Write(sync::Arc<parking_lot::Mutex<(evmap::WriteHandle<u64, u64>, usize, usize)>>),
 }
 
 impl Backend for EvHandle {
@@ -232,7 +240,7 @@ impl Backend for EvHandle {
 
     fn b_put(&mut self, key: u64, value: u64) {
         if let EvHandle::Write(ref w) = *self {
-            let mut w = w.lock().unwrap();
+            let mut w = w.lock();
             w.0.update(key, value);
             w.1 += 1;
             if w.1 == w.2 {
--- a/benchmark/src/main.rs
+++ b/benchmark/src/main.rs
@@ -88,7 +88,7 @@ fn main() {
     // first, benchmark Arc<RwLock<HashMap>>
     if matches.is_present("compare") {
         let map: HashMap<u64, u64> = HashMap::with_capacity(5_000_000);
-        let map = sync::Arc::new(sync::RwLock::new(map));
+        let map = sync::Arc::new(parking_lot::RwLock::new(map));
         let start = time::Instant::now();
         let end = start + dur;
         join.extend((0..readers).into_iter().map(|_| {
@@ -207,13 +207,13 @@ impl Backend for sync::Arc<CHashMap<u64, u64>> {
     }
 }
 
-impl Backend for sync::Arc<sync::RwLock<HashMap<u64, u64>>> {
+impl Backend for sync::Arc<parking_lot::RwLock<HashMap<u64, u64>>> {
     fn b_get(&self, key: u64) -> u64 {
-        self.read().unwrap().get(&key).map(|&v| v).unwrap_or(0)
+        self.read().get(&key).map(|&v| v).unwrap_or(0)
     }
 
     fn b_put(&mut self, key: u64, value: u64) {
-        self.write().unwrap().insert(key, value);
+        self.write().insert(key, value);
     }
 }
 
--- a/src/inner.rs
+++ b/src/inner.rs
@@ -1,17 +1,11 @@
+use std::hash::{BuildHasher, Hash};
+use std::sync::{atomic, Arc, Mutex};
+
 #[cfg(not(feature = "hashbrown"))]
 use std::collections::HashMap;
 
 #[cfg(feature = "hashbrown")]
 use hashbrown::HashMap;
-
-use std::hash::{BuildHasher, Hash};
-use std::sync::{atomic, Arc};
-
-#[cfg(not(feature = "parking_lot"))]
-use std::sync::Mutex;
-
-#[cfg(feature = "parking_lot")]
-use parking_lot::Mutex;
 
 #[cfg(not(feature = "smallvec"))]
 pub(crate) type Values<T> = Vec<T>;
--- a/src/inner.rs
+++ b/src/inner.rs
@@ -14,17 +14,17 @@ use std::sync::Mutex;
 use parking_lot::Mutex;
 
 #[cfg(not(feature = "smallvec"))]
-pub(crate) type Log<T> = Vec<T>;
+pub(crate) type Values<T> = Vec<T>;
 
 #[cfg(feature = "smallvec")]
-pub(crate) type Log<T> = smallvec::SmallVec<[T; 1]>;
+pub(crate) type Values<T> = smallvec::SmallVec<[T; 1]>;
 
 pub(crate) struct Inner<K, V, M, S>
 where
     K: Eq + Hash,
     S: BuildHasher,
 {
-    pub(crate) data: HashMap<K, Log<V>, S>,
+    pub(crate) data: HashMap<K, Values<V>, S>,
     pub(crate) epochs: Arc<Mutex<Vec<Arc<atomic::AtomicUsize>>>>,
     pub(crate) meta: M,
     ready: bool,
--- a/src/inner.rs
+++ b/src/inner.rs
@@ -1,13 +1,30 @@
+#[cfg(not(feature = "hashbrown"))]
 use std::collections::HashMap;
+
+#[cfg(feature = "hashbrown")]
+use hashbrown::HashMap;
+
 use std::hash::{BuildHasher, Hash};
-use std::sync::{atomic, Arc, Mutex};
+use std::sync::{atomic, Arc};
+
+#[cfg(not(feature = "parking_lot"))]
+use std::sync::Mutex;
+
+#[cfg(feature = "parking_lot")]
+use parking_lot::Mutex;
+
+#[cfg(not(feature = "smallvec"))]
+pub(crate) type Log<T> = Vec<T>;
+
+#[cfg(feature = "smallvec")]
+pub(crate) type Log<T> = smallvec::SmallVec<[T; 1]>;
 
 pub(crate) struct Inner<K, V, M, S>
 where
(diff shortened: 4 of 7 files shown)


benchmark/src/main.rs
```diff
@@ -4,6 +4,14 @@ extern crate clap;
...
 extern crate rand;
 extern crate zipf;
+extern crate parking_lot;
+
+#[cfg(all(not(target_env = "msvc"), feature = "nightly"))]
+extern crate jemallocator;
+
+#[cfg(all(not(target_env = "msvc"), feature = "nightly"))]
+#[global_allocator]
+static ALLOC: jemallocator::Jemalloc = jemallocator::Jemalloc;
```
jonhoo: This only improves the performance of the benchmarker, which I don't think should be a goal, so I'd cut this.

benchmark/src/main.rs
```diff
@@ -135,7 +143,7 @@ fn main() {
         let (r, w) = evmap::Options::default()
             .with_capacity(5_000_000)
             .construct();
-        let w = sync::Arc::new(sync::Mutex::new((w, 0, refresh)));
+        let w = sync::Arc::new(parking_lot::Mutex::new((w, 0, refresh)));
```
jonhoo: I wonder if `evmap` should provide this kind of shared write handle as a part of its API... What do you think?

src/inner.rs
```diff
@@ -1,13 +1,30 @@
...
+use std::sync::{atomic, Arc};
+
+#[cfg(not(feature = "parking_lot"))]
+use std::sync::Mutex;
+
+#[cfg(feature = "parking_lot")]
+use parking_lot::Mutex;
+
+#[cfg(not(feature = "smallvec"))]
+pub(crate) type Log<T> = Vec<T>;
```
jonhoo: I'd probably call this type `Values`, since it's not really a log.

Cargo.toml
```diff
@@ -21,3 +21,12 @@ travis-ci = { repository = "jonhoo/rust-evmap" }
 maintenance = { status = "passively-maintained" }
 
 [dependencies]
+hashbrown = { version = "0.1.8", optional = true }
+parking_lot = { version = "0.7.1", optional = true }
+smallvec = { version = "0.6.7", optional = true }
+
+[features]
+default = []
```
jonhoo: I think we'd probably want to include all the features by default.

src/lib.rs
```diff
@@ -184,7 +184,23 @@
...
+#[cfg(feature = "smallvec")]
+extern crate smallvec;
+
+/// Re-export default hash builder
+#[cfg(feature = "hashbrown")]
+pub type DefaultHashBuilder = hashbrown::hash_map::DefaultHashBuilder;
+
+/// Re-export default hash builder
+#[cfg(not(feature = "hashbrown"))]
+pub type DefaultHashBuilder = std::collections::hash_map::RandomState;
```
jonhoo: Hmm, I don't think this is okay, because Rust features [must be additive](https://github.com/rust-lang/cargo/issues/4323). In this case, imagine two crates, `a` and `b`, that both depend on `evmap`, but one does without the `hashbrown` feature, and one does with. Cargo will merge them into a single dependence on `evmap` with the feature enabled, at which point `a` won't compile since the types potentially won't match (if they explicitly name `Options<(), std::collections::hash_map::RandomState>` for example. I'm not entirely sure how to work around that though... Perhaps we just make `hashbrown` a non-optional dependency?

src/read.rs
```diff
@@ -59,9 +60,19 @@ where
 {
     // tell writer about our epoch tracker
     let epoch = sync::Arc::new(atomic::AtomicUsize::new(0));
-    inner.epochs.lock().unwrap().push(Arc::clone(&epoch));
+
+    {
```
jonhoo: This lock is taken so rarely that I don't know that it's worthwhile to turn it into a `parking_lot::Mutex`. The writer does take it on every `refresh`, but it should be entirely uncontended, and pretty much irrelevant. It'd be good to see a benchmark of this to see that it actually makes a difference before adding this complexity.

jonhoo:
Okay, so overall I'd say:

 - `parking_lot` is _probably_ not a worthwhile change based on my understanding of the current performance bottlenecks of `evmap`. A micro-benchmark would be good though. Specifically, the one `Mutex` in `evmap` is pretty much entirely uncontended, although it _is_ taken on every `refresh`. I don't know that `parking_lot::Mutex` improves on that case, but maybe?
 - `smallvec` is a *great* change.
 - `hashbrown` is tricky to add as an optional feature because it changes types that are part of `evmap`'s public type signatures, which in turn means that the features isn't [additive](https://github.com/rust-lang/cargo/issues/4323). I think we'd probably just want to make `hashbrown` non-optional if we believe it to be a pretty much universal win.
 - updates to the benchmark aren't particularly important, since they don't change the performance users will _experience_ when using `evmap` themselves. we may want `evmap` to expose a `MultiWriterHandle` for convenience though (which is essentially `Arc<Mutex<WriteHandle>>`).

benchmark/src/main.rs
novacrazy: I added this because there were massive performance regressions on nightly that could be explained by the switch from jemalloc to the system allocator. Oddly enough, it also affected the MSVC target, which I guess doesn’t have a fix like this.

jonhoo: Huh, how weird. Well, I still don't think it should be merged into `evmap` proper as I assume the underlying `nightly` issue will be fixed?

src/lib.rs
novacrazy: Hmm. Nothing really changes if they manually specify a hash builder, like with `Options::with_builder`.

Although not entirely additive, we could make it clear in the docs/reader that swapping out the standard hashmap for hashbrown does also swap between the default hash builder for both. 

Then again, with hashbrown being considered as a real replacement in the standard library, making it non-optional is also reasonable, but would be a breaking change. If the feature is left disabled right now the new type is still just `RandomState`.

benchmark/src/main.rs
novacrazy: I don’t know what else would cause it. I should note it affected the benchmark before I added anything, too, so whatever it is isn’t too relevant to this PR, but I included that to reduce future headaches. I can certainly remove it if you’d rather the benchmark stay agnostic to such things.

Cargo.toml
novacrazy: Perhaps the first two, but not `smallvec`. Of course `evmap` usually encourages small values like pointers, but with `smallvec` a physically large `V` would bloat up the hashmaps even when not used. There is also a tiny logical overhead of using `smallvec`, which is usually worth it if you don’t allocate, but a regular `Vec` would be better if you always do. So while in my project I would enable the feature, others may rather not.

benchmark/src/main.rs
novacrazy: I think the read/write handles make sense as-is, and there may even be situations where a mutex is unnecessary for the writer, e.g. some event loop that owns the writer and can use it mutably without a lock.

src/read.rs
novacrazy: Honestly, I still feel more comfortable with `parking_lot` if possible, given its additional safety features, such as no lock poisoning.

benchmark/src/main.rs
jonhoo: I was more thinking that we would provide this wrapper _in addition_ to the non-`Mutex` one so people don't have to do the wrapping themselves.

benchmark/src/main.rs
jonhoo: Yeah, I think I would. In particular, I'd like the benchmark to be somewhat representative of what people would likely get if they dropped `evmap` into their existing implementation, which doesn't quite hold if we require them to change the system allocator. Even if the system allocator is seemingly completely broken on some systems.

Cargo.toml
jonhoo: Yeah, that's a fair point. I do wonder whether this might lead to surprising results though. For example, imagine I am writing a crate that uses `evmap` and has some large `V`, and therefore intentionally does not enable `smallvec`. Then, someone uses my crate in a larger project that also depends on `evmap` _with_ `smallvec`. Cargo will resolve this situation by enabling the `smallvec` feature, and as a result, performance on my crate will be sad. This is a result of having this be a feature at all, not it being on/off by default though, so maybe that's just something we'll have to accept. But I do think it argues for enabling this by default, and add a warning to disable if you have large `V`, since the common case is likely that `V` is small.

src/lib.rs
jonhoo: No, the issue is if they use the default, but have written down the explicit type somewhere. Now, that said, doing so is probably rare if they're using the default random state, so maybe it's fine. As a somewhat long-shot, I'll ping @kennytm here to see if they have any thoughts, since they know far more about this than I do. I think I'm leaning towards saying that the change is fine as-is, though a comment in the docs is probably a worthwhile addition.

src/read.rs
jonhoo: Hmm, I'm not sure I consider no lock poisoning a safety feature? To the contrary, it means that you don't find out that the previous holder of the lock crashes, and thus may have corrupted the locked value's contents.

Cargo.toml
novacrazy: It wouldn't be *sad*, just a single extra branch. `smallvec` is most useful for it's memory locality. Combined with `hashbrown`, the entire map is basically a single flat array, which makes the CPU cache very happy.

jonhoo: Which I think speaks for `smallvec` + `hashbrown` being the default features. And (as discussed elsewhere), `hashbrown` not even being a feature.

src/read.rs
novacrazy: Almost all panics (aside from OOM errors) would be from user code, right? In that case, they would have to panic before an operation is added to the oplog or before a call to `refresh`. The oplog itself and all other members of the writer are uncorrupted, so the lock can be safely unlocked via RAII.

That would be an issue if they were adding many values in a loop, but only logically for their program, and they should probably handle that whenever their worker thread crashes, as it did not get a chance to finish adding all values. I mean, we can really only guarantee everything works up until the point it crashes. Guaranteeing eventual consistency still holds so long as their program can restart the worker thread and presumably *not* crash when adding the values again.

novacrazy: Actually, I just realized, the lock here is never even used in conjunction with user code, so it has zero chance of being poisoned anyway. I'm not sure if that's a point in favor or against keeping `parking_lot`.

jonhoo: I think that assumes that users aren't doing things that can panic in `Clone` for example. We clone in `refresh`, so if that panics for whatever reason, then later cloning a read handle wouldn't crash if you don't track poison. I think benchmarking this first is a good idea -- if the performance gains are negligible (which I think they'll be), then I think we should stick with the `std` lock.

src/lib.rs
jonhoo: Yeah, I'm also leaning towards 2, though would also be okay with 1. One option would be to do 1 under the assumption that the `std` `HashMap` will change to `hashbrown` in the future. Have you had a chance to benchmark the `hashbrown` swap in isolation to see that it actually speeds things up markedly?

novacrazy:
Running some new benchmarks now. I've limited the affinity to 8 real CPU cores (no SMT), and set it to high priority to hopefully reduce some noise.

jonhoo:
I think what you want to do is something similar to the script in [`benchmarks/README.md`](https://github.com/jonhoo/rust-evmap/tree/master/benchmark) and run with `#readers + #writers <= #cores`. That's where the interesting scalability results lie. Once you go beyond the number of cores, you end up context switching, and that's not particularly interesting from a performance standpoint.

novacrazy:
In that case, I'll bump it to 12 real cores and go overboard on the granularity of the benchmarks. This may take a while, but we should get some pretty graphs!

I've realized by now that these benchmarks are really just for people glancing at the library. Now that I understand how it works, the reader benchmarks are basically just benchmarking the hashmap implementation, plus a few atomic loads.

Speaking of... I just noticed an issue that I will fix after benchmarking.

In `ReadHandle::with_handle`, there is this snippet:

```rust
let rs = unsafe { Box::from_raw(r_handle) };
res = Some(f(&*rs));
mem::forget(rs); // don't free the Box!
```

However, if `f` panics (which can be in user code, via `get_and` and `for_each`, among others), `rs` **will** be dropped. Therefore, I will wrap it in [`ManuallyDrop`](https://doc.rust-lang.org/std/mem/struct.ManuallyDrop.html) ASAP, and look for other occurrences of this pattern.

jonhoo:
Yup, the benchmarker is essentially trying to generate as much contention as it can, because that's where we want to understand the behavior characteristics. If there's no contention, the performance is mostly just the performance of the underlying `HashMap` + some overhead. That's not all that interesting. Seeing how performance scales as you have more concurrent accesses however, that's where the real dragons lie! It's not about the sheer number of readers or writers, but rather about the access contention those readers and writers generate.

As for the snippet you gave, I think the real solution is to make that code simply:
```rust
res = Some(f(unsafe { &*handle }));
```
That way there's no issue with panics. I believe that should work just fine.

novacrazy:
I've been experiencing some odd results that don't reflect what I saw when first adding these features, to the point I'm beginning to think I'm running into memory bandwidth limitations on my system.

When I first benchmarked these, they ran at normal priority (or dynamically lower), and I had netflix on in the background. Typical stuff. It was a not-insignificant load, but it was consistent. I tried to not do anything drastic, but I didn't want to just stare at a blank screen for 20 minutes. That's when I observed the speedups.

However, when I increased the process priority and basically dedicated my system to the benchmarks, the results across all new features were equal, no change. 

We already agreed that the simple reader tests are basically just benchmarking the hashmap implementations, and I'm not even seeing the boosts that should be expected of hashbrown, given hashbrown's benchmarks destroy std's. 

So, what I think is going on is specific to me. Bad luck or bad planning, I suppose. As stated previously, I have a Threadripper 1950X, and I also have 128GB of RAM set up in a quad channel config. Unfortunately for 1st gen Threadripper, quad channel configs effectively halve the memory speed under certain workloads.

Randomly accessing super-optimized hashmaps at very high OS priorities appears to hit its limit.

What must have happened for the casual-use benchmarks is that it really **is** faster, and the OS was distributing the work in such a way that the presumed lower *operational latency* of operations with the new features gave it an edge. It was apparently still hitting that limit. However, the slower original config was having to be shared with the background tasks, and did not get a chance to hit that memory bandwidth limit.

This has certainly been an interesting adventure for so few lines of code changed.

What are your thoughts?

jonhoo:
Yeah, I've observed similar behavior in the past too. Its likely in part because the benchmark is so simple. One good strategy might be to run evmap through the [YCSB](https://research.yahoo.com/news/yahoo-cloud-serving-benchmark) or something else that's more realistic, and then measure the difference. There's also [this page](https://tessil.github.io/2016/08/29/benchmark-hopscotch-map.html) that measures hash tables directly, which we may be able to adapt to a concurrent workload. I also have a larger integration benchmark I could run with [noria](https://github.com/mit-pdos/noria) (cc @ms705), and that might help shine some light on the improvements.

Overall, I'd be inclined to merge this regardless; the additions are both sensible and optional. I do think, as discussed in one of the comments, that both features should be on by default though.

I'd also appreciate a fix to the `r_handle` drop issue you observed in a separate PR :)

novacrazy:
Perhaps we could use a slow CPU with fast memory for a fair benchmark? Or I may be able to underclock my CPU to compensate. 32 hardware threads at around 1 GHz (vs. 3.71 GHz now) would probably give enough contention. 

Anyway, I'll go ahead and make those optional features default and then it should be ready to merge.

src/lib.rs
novacrazy: After some thought, 2 would be the least breaking change, but we can re-export `hashbrown`'s FxHash implementation to make it obvious it exists. So I'm going to go with that.

novacrazy:
Everything should be ready.

benchmark/Cargo.toml
```diff
@@ -4,11 +4,18 @@ version = "0.1.0"
 authors = ["Jon Gjengset <jon@thesquareplanet.com>"]
 
 [dependencies]
-evmap = { path = "../" }
+evmap = { path = "../", features = ["hashbrown", "smallvec"] }
```
jonhoo: This can go away now I think?

benchmark/src/main.rs
jonhoo: If we're benchmarking evmap with `parking_lot`, we should also change the benchmark of the standard library hashmap to use `parking_lot`.

src/read.rs
```diff
@@ -59,9 +59,11 @@ where
 {
     // tell writer about our epoch tracker
     let epoch = sync::Arc::new(atomic::AtomicUsize::new(0));
+
```
jonhoo: Did `rustfmt` introduce this newline? If not, I'd prefer to keep it the way it was.

src/read.rs
```diff
@@ -59,9 +59,11 @@ where
...
     inner.epochs.lock().unwrap().push(Arc::clone(&epoch));
 
     let store = Box::into_raw(Box::new(inner));
+
```
jonhoo: ditto here

src/write.rs
```diff
@@ -117,7 +123,13 @@ where
         // destructors of any of the values that are in our map, as they'll all be called when the
         // last read handle goes out of scope.
         for (_, mut vs) in self.w_handle.as_mut().unwrap().data.drain() {
-            for v in vs.drain(..) {
+            #[cfg(not(feature = "smallvec"))]
+            let drain = vs.drain(..);
+
+            #[cfg(feature = "smallvec")]
+            let drain = vs.drain();
```
jonhoo: I wonder: to avoid potentially shooting ourselves in the leg later, would it work to do:
```rust
let drain;

#[cfg(not(feature = "smallvec"))]
drain = vs.drain(..);

#[cfg(feature = "smallvec")]
drain = vs.drain();
```
That way we can never accidentally call `.drain()` twice, because it'd be a double-assignment to a non-mutable variable. I _think_ that should compile, but if not, leave it as-is.

src/write.rs
```diff
@@ -181,7 +193,9 @@ where
         // only block on pre-existing readers, and they are never waiting to push onto epochs
         // unless they have finished reading.
         let epochs = Arc::clone(&self.w_handle.as_ref().unwrap().epochs);
+
```
jonhoo: ditto newlines

src/write.rs
```diff
@@ -181,7 +193,9 @@ where
...
         let mut epochs = epochs.lock().unwrap();
+
```
jonhoo: ditto newlines

src/write.rs
```diff
@@ -341,14 +356,17 @@ where
...
-        let r_handle = self.r_handle
+        let r_handle = self
+            .r_handle
             .inner
             .swap(ptr::null_mut(), atomic::Ordering::Release);
         let r_handle = unsafe { Box::from_raw(r_handle) };
 
         // now, wait for all readers to depart
         let epochs = Arc::clone(&self.w_handle.as_ref().unwrap().epochs);
+
```
jonhoo: ditto newlines

src/write.rs
```diff
@@ -341,14 +356,17 @@ where
...
         let mut epochs = epochs.lock().unwrap();
+
```
jonhoo: ditto newlines

src/write.rs
```diff
@@ -458,35 +476,59 @@ where
...
-                for v in vs.drain(..) {
-                    mem::forget(v);
+                let vs = inner.data.entry(key.clone()).or_insert_with(Values::new);
+
+                {
+                    #[cfg(not(feature = "smallvec"))]
+                    let drain = vs.drain(..);
+
+                    #[cfg(feature = "smallvec")]
+                    let drain = vs.drain();
```
jonhoo: Same not here about having `let drain;` first, and then do two non-`let` `drain = ` if possible.

src/write.rs
```diff
@@ -458,35 +476,59 @@ where
...
-                    .drain(..)
-                {
-                    mem::forget(v);
+                match inner.data.entry(key.clone()) {
+                    Entry::Occupied(mut occupied) => {
+                        #[cfg(not(feature = "smallvec"))]
+                        let drain = occupied.get_mut().drain(..);
+
+                        #[cfg(feature = "smallvec")]
+                        let drain = occupied.get_mut().drain();
```
jonhoo: ditto

src/write.rs
```diff
@@ -458,35 +476,59 @@ where
...
+                    .or_insert_with(Values::new)
                     .push(unsafe { value.shallow_copy() });
             }
             Operation::Empty(ref key) => {
                 if let Some(mut vs) = inner.data.remove(key) {
+                    #[cfg(not(feature = "smallvec"))]
+                    let drain = vs.drain(..);
+
+                    #[cfg(feature = "smallvec")]
+                    let drain = vs.drain();
```
jonhoo: ditto

src/read.rs
novacrazy: Indeed, it was rustfmt. I have it set up to automatically format on file save.

jonhoo: How weird -- I didn't expect it to force an extra line here. In that case, I'll resolve all of these!

src/write.rs
novacrazy: The two variations are mutually exclusive, so it doesn't really matter. If they were to somehow both exist, that would be a pretty major bug in the compiler.

benchmark/Cargo.toml
novacrazy: Probably, just to lessen the number of lines changed.

src/write.rs
jonhoo: I agree, but writing it as above ensures that we don't accidentally write `cfg`s later that are _not_ mutually exclusive :)

benchmark/src/main.rs
novacrazy: I thought about this, and I wanted the *least* overhead for the writers themselves. `parking_lot` is still faster than `std`, and we didn't want to benchmark the writer lock, only the concurrent map. However, for the `std` benchmark, the lock is part of that concurrent map, so it stayed on `std::sync::Mutex`.

I can revert this as well if you want.

src/write.rs
novacrazy: Well, upon trying it out, it seems we can't even do it that way.

```rust
let drain;

#[cfg(...)]
drain = whatever;
```

counts as an expression, and `#[cfg(...)]` is not yet supported for expressions, only declarations.

benchmark/src/main.rs
jonhoo: I'd argue that `parking_lot` *is* part of the concurrent map for `evmap` if you indeed want to have multiple write handles, so I think `std` should also get the advantage of using `parking_lot`. It won't materially change the results, but it is a fairer comparison.

src/write.rs
jonhoo: Bah, that's annoying. Oh well, keep it the way it is then :)

novacrazy:
I don't see whatever last pending review there is, but I think I covered everything.

jonhoo:
Awesome, thank you so much! And thanks for your patience in landing the PR too :D I'm excited to try this out in [noria](https://github.com/mit-pdos/noria) and see if we notice any speedup!

novacrazy:
Oh... you already released it? I still need to fix those `Drop` soundness issues first. You requested I make a separate PR for that after this one; was hoping to squeeze them into the same release.

PR #8, opened 2019-01-25
Fix a couple Drop soundness issues

novacrazy:
Fixes a couple issues where using a temporary `Box` to access a pointer could allow accidental drops.


jonhoo:
Perfect, thank you!

PR #9, opened 2019-01-26
Add more operations and API flexibility

novacrazy:
## Rationale 

For the application I intend to use `evmap` in, I keep an operational log of values with their own timestamps. `evmap` is ideal for this, as an operational log is built-in and natural. 

However, after a while, it becomes necessary to prune older versions of these values, and with the recently added `smallvec` feature it would be useful to occasionally renormalize the values to fit within the inline vector.

## Added Operations

To this end, I propose three new operations (or more if suggested):

* `Retain`
* `Fit`
* `Reserve`

`Retain` takes a closure predicate and converts it to a `'static` thread-safe trait object (`Box<Fn + Send>`), which can be used later to remove items from the value-set that do not satisfy the predicate. It works almost exactly like [`Vec::retain`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.retain)

`Fit` is simply short for `shrink_to_fit`, which for `smallvec` will move the item(s) back to the inline storage if possible and deallocate the heap storage. This is provided per-key or for the entire map.

`Reserve` pre-allocates space for a value-set, both in the hashmap and for the value-set elements themselves.

## Additional Thoughts

It would be nice to have a way to operate on the entire underlying map from user code, safely. One example would be an alternative `Retain` implementation that uses `rayon` to iterate over the entire map in parallel, or similar operations in parallel. Unfortunately, as of now I'm not sure how that could be done.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -190,10 +190,20 @@
 //! back to the inline storage if possible. This is ideal for maps that only ever use one
 //! element per key.
 //!
-//! If this is undesirable, simple set `default-features = false` in the `evmap` dependency entry,
-//! and `Vec` will always be used internally. Note that this will also opt out
-//! of the `hashbrown` dependency, which is usually preferred,
-//! so add that back with `features = ["hashbrown"]`
+//! If this is undesirable, simple set:
+//!
+//! ```toml
+//! default-features = false
+//! ```
+//!
+//! in the `evmap` dependency entry, and `Vec` will always be used internally.
+//!
+//! Note that this will also opt out of the `hashbrown` dependency, which is usually preferred,
+//! so add that back with:
+//!
+//! ```toml
+//! features = ["hashbrown"]
+//! ```
 //!
 #![deny(missing_docs)]
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -255,6 +255,12 @@ pub enum Operation<K, V> {
     Truncate(K, usize),
     /// Reverses the value-set for this key.
     Reverse(K),
+    /// Reserves capacity for some number of additional elements in a value-set,
+    /// or creates an empty value-set for this key with the given capacity if
+    /// it doesn't already exist.
+    ///
+    /// This can improve performance by pre-allocating space for large value-sets.
+    Reserve(K, usize),
 }
 
 mod write;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -182,6 +182,19 @@
 //! approximately twice of that of a regular `HashMap`, and more if writes rarely refresh after
 //! writing.
 //!
+//! # Small Vector Optimization
+//!
+//! By default, the value-set for each key in the map uses the `smallvec` crate to keep a
+//! maximum of one element stored inline with the map, as opposed to separately allocated
+//! with a plain `Vec`. Operations such as `Fit` and `Replace` will automatically switch
+//! back to the inline storage if possible. This is ideal for maps that only ever use one
+//! element per key.
+//!
+//! If this is undesirable, simple set `default-features = false` in the `evmap` dependency entry,
+//! and `Vec` will always be used internally. Note that this will also opt out
+//! of the `hashbrown` dependency, which is usually preferred,
+//! so add that back with `features = ["hashbrown"]`
+//!
 #![deny(missing_docs)]
 
 #[cfg(feature = "hashbrown")]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -195,13 +195,39 @@ extern crate smallvec;
 pub type FxHashBuilder = hashbrown::hash_map::DefaultHashBuilder;
 
 use std::collections::hash_map::RandomState;
+use std::fmt;
 use std::hash::{BuildHasher, Hash};
 
 mod inner;
 use inner::Inner;
 
+/// Unary predicate used to retain elements
+pub struct Predicate<V>(pub Box<Fn(&V) -> bool + Send>);
+
+impl<V> Predicate<V> {
+    /// Evaluate the predicate for the given element
+    #[inline]
+    pub fn eval(&self, value: &V) -> bool {
+        (*self.0)(value)
+    }
+}
+
+impl<V> PartialEq for Predicate<V> {
+    fn eq(&self, _: &Self) -> bool {
+        // the predicate is uniquely owned,
+        // so they will never be equivalent.
+        false
+    }
+}
+
+impl<V> fmt::Debug for Predicate<V> {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "Predicate({:p})", &*self.0 as *const _)
+    }
+}
+
 /// A pending map operation.
-#[derive(Clone, PartialEq, Eq, Debug)]
+#[derive(Debug, PartialEq)]
 pub enum Operation<K, V> {
     /// Replace the set of entries for this key with this value.
     Replace(K, V),
@@ -213,6 +239,17 @@ pub enum Operation<K, V> {
     Empty(K),
     /// Remove all values in the value set for this key.
     Clear(K),
+    /// Retains all values matching the given predicate.
+    Retain(K, Predicate<V>),
+    /// Shrinks a value-set to it's minimum necessary size, freeing memory
+    /// and potentially improving cache locality if the `smallvec` feature is used.
+    ///
+    /// If no key is given, all value-sets will shrink to fit.
+    Fit(Option<K>),
+    /// Truncates the value-set for this key to the given length.
+    Truncate(K, usize),
+    /// Reverses the value-set for this key.
+    Reverse(K),
 }
 
 mod write;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -197,12 +197,14 @@ pub type FxHashBuilder = hashbrown::hash_map::DefaultHashBuilder;
 use std::collections::hash_map::RandomState;
 use std::fmt;
 use std::hash::{BuildHasher, Hash};
+use std::sync::Arc;
 
 mod inner;
 use inner::Inner;
 
 /// Unary predicate used to retain elements
-pub struct Predicate<V>(pub Box<Fn(&V) -> bool + Send>);
+#[derive(Clone)]
+pub struct Predicate<V>(pub(crate) Arc<Fn(&V) -> bool + Send + Sync>);
 
 impl<V> Predicate<V> {
     /// Evaluate the predicate for the given element
@@ -213,12 +215,13 @@ impl<V> Predicate<V> {
 }
 
 impl<V> PartialEq for Predicate<V> {
-    fn eq(&self, _: &Self) -> bool {
-        // the predicate is uniquely owned,
-        // so they will never be equivalent.
-        false
+    #[inline]
+    fn eq(&self, other: &Self) -> bool {
+        Arc::ptr_eq(&self.0, &other.0)
     }
 }
+
+impl<V> Eq for Predicate<V> {}
 
 impl<V> fmt::Debug for Predicate<V> {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
@@ -227,7 +230,7 @@ impl<V> fmt::Debug for Predicate<V> {
 }
 
 /// A pending map operation.
-#[derive(Debug, PartialEq)]
+#[derive(Clone, PartialEq, Eq, Debug)]
 pub enum Operation<K, V> {
     /// Replace the set of entries for this key with this value.
     Replace(K, V),
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -225,7 +225,9 @@ impl<V> Eq for Predicate<V> {}
 
 impl<V> fmt::Debug for Predicate<V> {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "Predicate({:p})", &*self.0 as *const _)
+        f.debug_tuple("Predicate")
+            .field(&format_args!("{:p}", &*self.0 as *const _))
+            .finish()
     }
 }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -185,10 +185,10 @@
 //! # Small Vector Optimization
 //!
 //! By default, the value-set for each key in the map uses the `smallvec` crate to keep a
-//! maximum of one element stored inline with the map, as opposed to separately allocated
+//! maximum of one element stored inline with the map, as opposed to separately heap-allocated
 //! with a plain `Vec`. Operations such as `Fit` and `Replace` will automatically switch
-//! back to the inline storage if possible. This is ideal for maps that only ever use one
-//! element per key.
+//! back to the inline storage if possible. This is ideal for maps that mostly use one
+//! element per key, as it can improvate memory locality with less indirection.
 //!
 //! If this is undesirable, simple set:
 //!
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -251,10 +251,6 @@ pub enum Operation<K, V> {
     ///
     /// If no key is given, all value-sets will shrink to fit.
     Fit(Option<K>),
-    /// Truncates the value-set for this key to the given length.
-    Truncate(K, usize),
-    /// Reverses the value-set for this key.
-    Reverse(K),
     /// Reserves capacity for some number of additional elements in a value-set,
     /// or creates an empty value-set for this key with the given capacity if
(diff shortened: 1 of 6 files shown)


novacrazy:
I'm also looking into optimizing how the oplog is evaluated to avoid repeated lookups for consecutive operations.

src/lib.rs
```diff
@@ -195,11 +195,42 @@ extern crate smallvec;
...
+use std::fmt;
 use std::hash::{BuildHasher, Hash};
+use std::sync::Arc;
 
 mod inner;
 use inner::Inner;
 
+/// Unary predicate used to retain elements
+#[derive(Clone)]
+pub struct Predicate<V>(pub(crate) Arc<Fn(&V) -> bool + Send + Sync>);
```
jonhoo: I _think_ it should be possible to make this a `Box` instead of an `Arc`?

src/lib.rs
```diff
@@ -213,6 +244,23 @@ pub enum Operation<K, V> {
...
+    Retain(K, Predicate<V>),
+    /// Shrinks a value-set to it's minimum necessary size, freeing memory
+    /// and potentially improving cache locality if the `smallvec` feature is used.
+    ///
+    /// If no key is given, all value-sets will shrink to fit.
+    Fit(Option<K>),
+    /// Truncates the value-set for this key to the given length.
+    Truncate(K, usize),
+    /// Reverses the value-set for this key.
+    Reverse(K),
```
jonhoo: `Reverse` and `Truncate` strike me as odd. I sort of see how they're analogous to the operations you can do on regular Rust collections, but for a value-set they seem somewhat strange. I wonder if we want to provide more "semantic" operations like `Pick(Front|Back, usize)`?

src/lib.rs
novacrazy: I had it as a `Box` earlier, but it required removing `Eq` and `Clone` from the `Operation` enum. It doesn't add any overhead to calling the function, but it is cloneable.

src/write.rs
```diff
@@ -222,15 +222,26 @@ where
...
                 // ourselves a lot of re-hashing, re-bucketization, etc.
                 w_handle
                     .data
                     .extend(r_handle.data.iter_mut().map(|(k, vs)| {
-                        (
-                            k.clone(),
-                            vs.iter_mut().map(|v| unsafe { v.shallow_copy() }).collect(),
-                        )
+                        (k.clone(), {
+                            // pre-allocate if possible
```
jonhoo: I believe `collect()` will already do that if the iterator is `ExactSizeIterator`, which `iter_mut` already provides, so I don't think this will actually have any effect.

src/write.rs
```diff
@@ -222,15 +222,26 @@ where
                         .unwrap()
                 };
 
+                let free_elements = w_handle.data.capacity() - w_handle.data.len();
+
+                if free_elements < r_handle.data.len() {
+                    // pre-allocate if possible
+                    w_handle.data.reserve(r_handle.data.len() - free_elements);
+                }
```
jonhoo: I believe (from the comment above) that `w_handle.data` should be empty here, which means that we can probably do better than this. Specifically, I think below we can replace
```rust
w_handle.data.extend(r_handle.data.iter_mut().map(...))
```
with
```rust
mem::replace(&mut w_handle.data, r_handle.data.iter_mut().map(...).collect())
```
It'd be good to assert that the returned `.data` is indeed empty there though, and then run the test suite.

jonhoo:
I think I'd like to see the `-> &mut Self` change in its own PR if possible?

src/write.rs
```diff
@@ -548,9 +611,61 @@ where
                     }
                 }
             }
+            Operation::Retain(ref key, ref predicate) => {
+                if let Some(e) = inner.data.get_mut(key) {
+                    let mut del = 0;
+
```
jonhoo: There are an awful lot of empty lines here that I don't think `rustfmt` would add automatically. I generally prefer to use empty lines to denote groups of logically connected statements, as opposed to using an extra newline between every line.

src/write.rs
```diff
@@ -548,9 +611,61 @@ where
...
+                    let len = e.len();
+
+                    for i in 0..len {
+                        if !predicate.eval(unsafe { e.get_unchecked(i) }) {
+                            del += 1;
+                        } else {
+                            e.swap(i - del, i);
+                        }
+                    }
```
jonhoo: This could use a comment describing what's going on. You're basically moving all the things we're dropping to the end, and then forgetting them, right? Writing that out would be helpful.

src/write.rs
```diff
@@ -548,9 +611,61 @@ where
...
+
+                    // manually truncate and forget values so they aren't dropped.
+                    while e.len() > (len - del) {
```
jonhoo: I think we can just truncate with `e.set_len()` here?

src/write.rs
```diff
@@ -548,9 +611,61 @@ where
...
+                }
+                None => {
+                    for value_set in inner.data.values_mut() {
+                        value_set.shrink_to_fit();
+                    }
+                }
+            },
+            Operation::Truncate(ref key, len) => {
+                if let Some(e) = inner.data.get_mut(key) {
+                    while e.len() > len {
```
jonhoo: Same: `e.set_len()`

src/write.rs
novacrazy: Yeah, it's "bubbling up" the elements we wish to remove, then popping them all off at the end. This was almost directly copied from `smallvec`'s `retain` implementation.

src/write.rs
novacrazy: That is a great idea!

src/lib.rs
jonhoo: Hmm, yeah, I suppose we sort of want to use `shallow_copy` for this `Fn` too, but that's trickier to achieve. `Arc` is probably fine.

src/write.rs
jonhoo: I think it would be helpful to add a comment + maybe even linking to `Vec::retain` and then say that we "do pretty much the same", but have to do it manually so that we don't drop anything

src/write.rs
jonhoo: An empty `std` collection always consumes no memory in Rust, so I don't think that's true?

jonhoo: Also, I'm pretty sure `extend` also uses the pre-allocate on `ExactSizeIterator` trick, so this may all be unimportant.

src/write.rs
jonhoo: I think linking to `Vec::retain` and saying in the same place that we can't use `.retain` because of dropping would be a good change. Makes the code easier to read for someone who's not familiar with it!

src/write.rs
novacrazy: Only for a fresh new one, but one that previously had elements but they were all removed still keeps that allocated memory around. It would be more accurate to say they are lazily allocated.

jonhoo: True, but in this case, unless I'm misremembering, they should truly be empty (as in, freshly allocated). See the comment above in the code.

novacrazy: Ah, I see what you mean. Still, it doesn't really matter. `collect()` will simply call `extend` internally.

jonhoo: Yup, which in turn uses the `ExactSizeIterator` to pre-reserve when possible.

src/lib.rs
novacrazy: After some thought I agree with you. We keep using the term "value-set", so perhaps we should restrict operations to those found in sets, like `HashSet`? More or less.

I mostly added `Reverse` and `Truncate` because they were easy and simple to expose, but they aren't necessary. It's probably best to not include those unless it becomes necessary. Reverse by itself especially isn't necessary, if the reader simply iterates in reverse.

jonhoo: Okay, yeah, let's remove `Reverse` and `Truncate` for the time being then, and instead add them if there should be demand for them later?

tests/lib.rs
```diff
@@ -387,3 +387,30 @@ fn foreach() {
...
+    let mut v = Vec::new();
+    let (r, mut w) = evmap::new();
+
+    for i in 0..50 {
+        w.insert(0, i);
+        v.push(i);
+    }
+
+    w.retain(0, |num| num % 2 == 0).refresh();
+    v.retain(|num| num % 2 == 0);
```
jonhoo: I think I'd factor out the closure here since the whole point is that they should be exactly the same.

tests/lib.rs
```diff
@@ -387,3 +387,30 @@ fn foreach() {
...
+
+    r.get_and(&0, |nums| {
+        assert_eq!(nums.len(), 25);
```
jonhoo: I would remove this line, as the line below should do the right thing on its own.

tests/lib.rs
```diff
@@ -387,3 +387,30 @@ fn foreach() {
         _ => unreachable!(),
     });
 }
+
+#[test]
+fn retain() {
```
jonhoo: I'd like to also see tests for `Fit` and `Reserve`, though I guess that might be hard? Having a test that at least _runs_ the operations and then do an insert+get to see that it doesn't crash would be good though.

tests/lib.rs
```diff
@@ -387,3 +387,30 @@ fn foreach() {
...
+        w.insert(0, i);
+        v.push(i);
+    }
+
+    w.retain(0, |num| num % 2 == 0).refresh();
+    v.retain(|num| num % 2 == 0);
+
+    r.get_and(&0, |nums| {
+        assert_eq!(nums.len(), 25);
+        assert_eq!(nums.len(), v.len());
```
jonhoo: I wonder if I would just
```rust
assert_eq!(nums, v);
```
and cut the check below.

src/lib.rs
novacrazy: Somehow I didn't see this. 

Shallow copy isn't needed for this at all. This doesn't carry any value nor is it inserted into the map itself. It's simply a callback used to evaluate the `retain` operation.

novacrazy:
For `smallvec`, we had previously discussed how it most helped with many reads of a single value, with mostly ever `replace` operations.

To that end, I've made `replace` automatically call `shrink_to_fit` when using `smallvec`, before pushing the replacement value. This forces it to switch back to the inline storage if it had been heap allocated before, or is a no-op if it was already on the inline storage.

The worst case scenario for this behavior would be more than one `Add`/`insert` operation, followed by a single replace. That seems rather unlikely, and if they are doing something that odd they should probably remove the `smallvec` feature and use plain `Vec`s.

Of course, a more natural pattern for the above scenario would be to never use `replace` at all, and instead use only `insert` and `clear`. In which case, this change has no affect, and they can manually call `fit` as necessary.

Therefore, I think this is a safe change that will help automatically renormalize the `SmallVec`s for applications where they do occasionally have many values, but it's usually a single value.

src/read.rs
```diff
@@ -255,3 +267,29 @@ where
...
+mod test {
+    use crate::new;
+
+    #[test]
+    fn reserve_and_fit() {
+        let (r, mut w) = new();
+
+        // 64 is the second power of two higher than 17,
+        // so it should at least shrink the vector to less than
+        // or equal to the next highest power of 2 after 17
```
jonhoo: Where does 17 come from?

jonhoo: Ohh, because of how many things we insert further down, I see..

jonhoo:
@novacrazy Yeah, I think you're totally right about `.set_len(0)` vs `drain` + `forget`. Good catch!

I'm of two minds when it comes to `shrink_to_fit` after replace when using `smallvec`. My concern is that this only makes sense if your workload is such that there are very occasionally many values for a key, but then many operations where there's only one. We basically have these cases:

 - A given value-set generally has many values: the `shrink_to_fit` will cause unnecessary copies and re-allocations.
 - A given value-set only has a single value: the `shrink_and_fit` has no effect.
 - A given value-set _occasionally_ has many values: the `shrink_to_fit` improves performance for the periods when there is only one value.

I don't know which of these are more common. I wonder whether the most reasonable approach is to _not_ `shrink_to_fit` by default, and then mention in the docs under `smallvec` that _if_ you know the value-set will have one value going forward after potentially having multiple values, then you should explicitly call `Fit`.

novacrazy:
A more idiomatic approach to many values would be to not use `replace` at all, but rather using `clear` and `insert`. So perhaps it would best be mentioned in the documentation for `replace` that such a workload be organized that way?

Furthermore, `shrink_to_fit` while the vector is empty does not copy any data, versus calling `fit` *afterwards* would require both another key lookup and copying of the value. In fact, LLVM may be smart enough to see we set the length to 0 immediately prior and optimize away the zero-length copying instructions entirely.

src/read.rs
jonhoo: I think it'd be useful to add something like
```rust
// 1 + 2^4, so it's a tiny bit more than a power of two
let n = 17;
```
at the top of the test, and then use `n` instead of a literal `17` elsewhere.

jonhoo:
Mmm, that's a good point. I'd add that to the docs of `replace` then, and then keep `shrink_and_fit`. Notice that that _does_ mean that `clear()` should _not_ `shrink_and_fit`, and that should also be documented.

novacrazy:
That has a good symmetry with smallvec’s documentation, too, which explicitly states its `clear` will not `shrink_to_fit`.

jonhoo:
Awesome! Thanks for all your hard work on this!

novacrazy:
Of course! Are there any more operations you can think of that would be nice to have?

If you don't have any, I'll probably squash and merge this today.

However, there are three more things on my radar right now, that fit better in separate PRs:

1. Optional serde support.
    * The implementation itself is easy, but the details of how it should be exposed are more complicated. 
    * Your statement on my first PR about including some sort of `Arc<Mutex<WriteHandle>>` in the crate itself could be useful for this, by providing a bundled `ReadWriteHandle`, which could trivially and safely implement `Serialize` and `Deserialize`, and only `Serialize` for any `ReadHandle`/`WriteHandle`
2. Provide a way to execute many mostly-arbitrary operations on a single value-set in a drop-agnostic way, probably via some kind of wrapper. 
    * An enum could work, with a tiny bit of overhead, but it would still be less overhead than many individual consecutive operations. I'm still exploring possibilities.
    * It would be run similarly to `Retain`, where an op is queued with an arbitrary user-given callback, which would be provided with that drop-agnostic structure with all ops exposed on it.

e.g.:

```rust
w.combined(k, |vs: SomeWrapper| {
    vs.clear().
      .insert(10)
      .insert(20)
      .fit();
}).refresh();
```

3. Avoid repeated table lookups/key hashes for consecutive operations on the same key/value-set. 
    * This could possibly be done with the raw-entry API on nightly, or stable with hashbrown.
    * It may even be useful to scan ahead for same-key ops and effectively merge them all together.

1 is simply a QoL thing, and 2/3 are purely writer optimization, so they aren't wholly necessary, but every little bit helps.

jonhoo:
I think I'd prefer to err on the side of adding few operations and then extending the list as use-cases for them arise, so I'd say go ahead and squash+merge!

As for the other features:

 1. Serialization support is actually not entirely straightforward, because it requires the ability to enumerate the entire data-structure. Maybe we could add support for it just through a `&mut WriteHandle`? We _could_ allow serializing through a `ReadHandle`, though that would require holding up any writes until the entire walk of the map has been completed. Though maybe that's fine with enough of a disclaimer?

    As for `Arc<Mutex<WriteHandle>>`, it's worth keeping in mind that you can already read through a `WriteHandle`. The `Arc<Mutex>` wrapper would only really be there to allow a simple if inefficient interface to have multiple writers. For deserializing, I think we need something different like deserializing into a `(WriteHandle, ReadHandle)`. Alternatively, we could deserialize into just a `WriteHandle` and make it easier to construct a `ReadHandle` from a `WriteHandle` (you already can by derefing and then cloning, but it's a bit awkward).

 2. What makes you say that a multi-op operation would be more efficient? I guess it'll save some key lookups perhaps, but we _could_ improve that just by remembering the previous operation's entry and checking if the next is for the same key. An intermediate would be to remember the hashes and use the new raw entry API. I think a multi-op would only really be useful if it allowed you to also read things from the entry while you're in the closure, so you can effectively do something that amounts to single-key transactions. Could perhaps even be expanded to multi-key, but that's a stretch for now.

 3. Haha, I guess we thought exactly alike since that's what I suggested above. I think that's a great idea. Per-key collapsing _should_ also be fine I think, though it's unclear to me how much we could actually do and have it be worth it.

novacrazy:
It may be a while until I can work out those things I listed, so it's probably okay to release a version `4.2.0` with these improvements. How does that sound?

PR #13, opened 2019-02-07
WIP: Operation lookahead and merging, and more complex nested modifier operations.

novacrazy:
This PR does the following:

  1. Splits operations into two types: `MapOperation` and `ValueOperation`, and uses them within a single parent `Operation` enum like so:
     
     ```rust
     pub enum Operation<K, V> {
         Value { key: K, op: ValueOperation<V> },
         Map { op: MapOperation<V> }.
     }
     ```
     
     splitting the operation types in this way allows for easy access to the key value, which is important for:

  2. Cache key hashes and look ahead in the operation log to eagerly apply operations for same-key value-sets without having to rehash every time.
     
     For keys with non-trivial hashes or complex hashing algorithms, this can avoid much unnecessary rehashing. 
     
     This uses the `hashbrown` Raw Entry API, which does exist in `std::collections::hash_map::HashMap`, but is stuck on nightly Rust for now. Therefore, hashbrown is now a non-optional feature.
     
Furthermore, a new kind of operation was introduced:

  3. Arbitrary `Modify` and `ForEach` operations which are allowed to the the (potentially aliased) data in the value-sets *while* applying modifications to them.
     
     Example:
     
     ```rust
     w.modify(some_key, |vs| {
         let first = {
             let vs: &[i32] = unsafe { vs.get() };
     
             vs.first().unwrap_or(0);
         };
     
         if first < 10 {
             vs.replace(20);
         }
     
         vs.insert(10).insert(11).insert(12).fit();
     });
     ```

     This gives much more flexibility. It has been made `unsafe` to access the existing value-set, as it could be accessing potentially aliased data. This use of `unsafe` is simply a warning flag for the user, and a note has been made in the documentation that no attempt to clone these values should be made. It is purely for those very rare cases where such dynamic logic is necessary.
     
     However, even without using the above, this provides a way to apply many operations to a single value-set with almost no overhead whatsoever. It requires only a single key hash and single table lookup.

  4. Adds a few extra operations in addition to `Modify`/`ForEach`:
     
     * `MapOperation::Prune`, which removes empty value-sets from the map.
     * `ValueOperation::RemoveStable`, which removes a single value from a value-set without affecting the order of elements, at the cost of being slower.
         * This is opposed to `ValueOperation::Remove`, which is faster but can swap around elements, affecting the order. 

## Implementation Details

In order to implement the lookahead in such a way that works both with consuming contexts and non-consuming shallow-copy contexts, operations had to be wrapped in a type that can sufficiently track this and allow for interactions between them.

This is done with `MarkedOperation`, implemented like so:

```rust
pub(crate) struct MarkedOperation<K, V> {
    pub op: mem::ManuallyDrop<UnsafeCell<Operation<K, V>>>,
    pub flag: Cell<u32>,
}

const NONE_FLAG: u32 = 0; // value has not been used at all
const FIRST_FLAG: u32 = 1; // value has been used without consuming
const SECOND_FLAG: u32 = 2; // value has been consumed
```

with operations such as:

<details>
<summary>click here to see implementation code</summary>

```rust
impl<K, V> Drop for MarkedOperation<K, V> {
    fn drop(&mut self) {
        // If the operation hasn't been consumed, drop it
        match self.flag.get() {
            NONE_FLAG | FIRST_FLAG => unsafe {
                mem::ManuallyDrop::drop(&mut self.op);
            },
            _ => {}
        }
    }
}

impl<K, V> MarkedOperation<K, V> {
    #[inline(always)]
    pub fn as_ref(&self) -> Option<&Operation<K, V>> {
        match self.flag.get() {
            NONE_FLAG => Some(unsafe { &*self.op.get() }),
            _ => None,
        }
    }

    #[inline(always)]
    pub fn take_second(&self) -> Option<Operation<K, V>> {
        match self.flag.get() {
            SECOND_FLAG => None,
            _ => {
                self.flag.set(SECOND_FLAG);

                Some(unsafe { ptr::read(self.op.get()) })
            }
        }
    }
}
```
</details>

While those may have unsafe code in them, the flag prevents any undefined behavior from occurring, and allowing for a more complex lifecycle for the internal value, with zero overhead besides the flag itself.

While it would be technically possible to emulate this with `Option`s and a boolean, that is effectively splitting the `flag` into two variables, and adding unnecessary overhead both in the compiled code and the number of nested `match` statements that must be used. It gets pretty messy, up to two or three match statements doing the exact same thing, just to get to the inner values. With `MarkedOperation`, it is a single `match` statement. 

All `unsafe` code is sufficiently compartmentalized and accounted for.

## In summary

I think this will have significant impact with applications that use complex keys, and provide the necessary flexibility I've been needing for my own applications. Even for simple integer keys, the low additional overhead provided by `MarkedOperation` and the lookahead should be negligible. 

In-depth benchmarks are planned, certainly. 

A quick test with Noria (of which I'm not too familiar with) yielded around 20%-40% latency improvements, e.g. 50000µs down to 30000µs for 100% read/write results on the "vote" benchmark.

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/rust-evmap/13)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,10 +21,9 @@ travis-ci = { repository = "jonhoo/rust-evmap" }
 maintenance = { status = "passively-maintained" }
 
 [dependencies]
-hashbrown = { version = "0.1.8", optional = true }
-smallvec = { version = "0.6.7", optional = true }
+hashbrown = { version = "0.1.8" }
+smallvec = { version = "0.6.7" }
 
 [features]
-default = ["hashbrown", "smallvec"]
 nightly_hashbrown = ["hashbrown/nightly"]
 nightly_smallvec = ["smallvec/union", "smallvec/may_dangle", "smallvec/specialization"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,5 +25,6 @@ hashbrown = { version = "0.1.8" }
 smallvec = { version = "0.6.7", optional = true }
 
 [features]
+default = ["smallvec"]
 nightly = ["hashbrown/nightly"]
 nightly_smallvec = ["smallvec/union", "smallvec/may_dangle", "smallvec/specialization"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,8 @@ maintenance = { status = "passively-maintained" }
 
 [dependencies]
 hashbrown = { version = "0.1.8" }
-smallvec = { version = "0.6.7" }
+smallvec = { version = "0.6.7", optional = true }
 
 [features]
-nightly = ["hashbrown/nightly", "smallvec/union", "smallvec/may_dangle", "smallvec/specialization"]
+nightly = ["hashbrown/nightly"]
+nightly_smallvec = ["smallvec/union", "smallvec/may_dangle", "smallvec/specialization"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,5 +25,4 @@ hashbrown = { version = "0.1.8" }
 smallvec = { version = "0.6.7" }
 
 [features]
-nightly_hashbrown = ["hashbrown/nightly"]
-nightly_smallvec = ["smallvec/union", "smallvec/may_dangle", "smallvec/specialization"]
+nightly = ["hashbrown/nightly", "smallvec/union", "smallvec/may_dangle", "smallvec/specialization"]
--- a/src/inner.rs
+++ b/src/inner.rs
@@ -1,16 +1,8 @@
 use std::hash::{BuildHasher, Hash};
 use std::sync::{atomic, Arc, Mutex};
 
-#[cfg(not(feature = "hashbrown"))]
-use std::collections::HashMap;
-
-#[cfg(feature = "hashbrown")]
 use hashbrown::HashMap;
 
-#[cfg(not(feature = "smallvec"))]
-pub(crate) type Values<T> = Vec<T>;
-
-#[cfg(feature = "smallvec")]
 pub(crate) type Values<T> = smallvec::SmallVec<[T; 1]>;
 
 pub(crate) struct Inner<K, V, M, S>
--- a/src/inner.rs
+++ b/src/inner.rs
@@ -3,7 +3,11 @@ use std::sync::{atomic, Arc, Mutex};
 
 use hashbrown::HashMap;
 
+#[cfg(feature = "smallvec")]
 pub(crate) type Values<T> = smallvec::SmallVec<[T; 1]>;
+
+#[cfg(not(feature = "smallvec"))]
+pub(crate) type Values<T> = Vec<T>;
 
 pub(crate) struct Inner<K, V, M, S>
 where
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -186,6 +186,8 @@
 #![deny(missing_docs)]
 
 extern crate hashbrown;
+
+#[cfg(feature = "smallvec")]
 extern crate smallvec;
 
 /// Re-export default FxHash hash builder from `hashbrown`
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -183,8 +183,7 @@
 //! writing.
 //!
 
-//#![deny(missing_docs)]
-#![allow(dead_code)]
+#![deny(missing_docs)]
 
 extern crate hashbrown;
 extern crate smallvec;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -182,105 +182,24 @@
 //! approximately twice of that of a regular `HashMap`, and more if writes rarely refresh after
 //! writing.
 //!
-//! # Small Vector Optimization
-//!
-//! By default, the value-set for each key in the map uses the `smallvec` crate to keep a
-//! maximum of one element stored inline with the map, as opposed to separately heap-allocated
-//! with a plain `Vec`. Operations such as `Fit` and `Replace` will automatically switch
-//! back to the inline storage if possible. This is ideal for maps that mostly use one
-//! element per key, as it can improvate memory locality with less indirection.
-//!
-//! If this is undesirable, simple set:
-//!
-//! ```toml
-//! default-features = false
-//! ```
-//!
-//! in the `evmap` dependency entry, and `Vec` will always be used internally.
-//!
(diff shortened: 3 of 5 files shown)


jonhoo:
This is awesome! I'll try to give the PR a proper read-through in the next few days, but a few high-level comments:

 - I'm not sure I follow the logic for why accessing the data in a value-set has to be `unsafe`? Even though the data is aliased, giving out `&` access to it should be fine (`&` tolerates aliasing after all). Cloning the underlying data should also be fine I think?
 - I _really_ want `MarkedOperation` to be replaced with a fully safe variant along the lines of:
   ```rust
   pub(crate) struct MarkedOperation<K, V> {
       pub op: Option<Operation<K, V>>,
       pub applied: bool,
   }
   ```
   This does not increase the number of fields or variables, and avoids all `unsafe`. It _will_ require that we give `apply_*` mutable access to the oplog, but I don't think that should be hard, since `apply_*` is always executed in the context of a writer, which as `&mut self` anyway. It might require you to iterate over indices instead of values in the outer loop to allow mutable access to subsequent elements in the look-ahead application, but I think that's worth it. `slice::split_at_mut` may come in handy.
 - Why does this PR make `smallvec` non-optional? As far as I can tell, there isn't a reason to do so?
 - I wonder if `Predicate` and `Modifier` can use `Box` instead of `Arc`? They should only be present in the oplog once iiuc? And at that point I think they can also be `FnMut` once we give `apply_*` mutable access to the oplog.
 - I'd like to see some more unit tests being added both to test these operations in isolation, and to test things like the oplog containing a mix of global and key-local operations. And in particular, there should probably be several tests that check that the look-ahead works correctly in the presence of multiple operations and `refresh()` calls.

novacrazy:
Apologies in advance if my response is a bit uncoordinated. Head is quite fuzzy today.

Simple things first:

* `smallvec` could be made optional again. No issue with that, I think.

* `Predicate` and `Modifier` require `Arc` because of the requirements for `Clone`, `PartialEq` and `Eq` on `Operation`. If the `Clone` requirement were to be removed, it would be possible, and `PartialEq`/`Eq` would simply always be false due to the unique ownership.

* More unit tests and benchmarks are definitely planned. If you have any ideas I'd be happy to add them. Everything current seems to work, at least.

* As for why accessing the data in a `Modify` operation is unsafe... the modifiers act on the `w_handle` directly, not the `r_handle`, and can therefore have technically-valid but actually invalid copies of data (I think), such as shallow-copied `Arc`s. Would those not have the correct number of references? It's entirely possible I'm misunderstanding that aspect of the operation evaluation. 

For the hard part, here is my train of thought: `Option` is a plain Rust enum, and can therefore be represented [(more or less)](https://doc.rust-lang.org/reference/items/unions.html) as:

```rust
union OptionData<T> {
    some: T,
    none: (),
}

struct Option<T> {
    data: OptionData<T>,
    discriminator: u32, // or just u8
}
```

with a few `unsafe` operations to correctly retrieve whichever value. Normally, Rust handles all this under the hood safely and automatically.

However, for `MarkedOperation` as I have it, all I've really done is strip away the extra wrapper and combined the `applied` and `discriminator` values into a single flag. I suppose that is reinventing the wheel a little, but it's no less safe. 

`union`s in Rust naturally cannot implement `Drop`, and in fact `std::mem::ManuallyDrop` is literally just a `union` with a single value. 

```rust
pub union ManuallyDrop<T>{ value: T }
```

It's totally transparent, with the added benefit of it not automatically dropping values. 

`UnsafeCell`, despite its name, is simply defined as:

```rust
struct UnsafeCell<T> {
    value: T,
}
```

and all it does is provide a mutable pointer to the data it holds. It's "unsafe" only when that pointer is dereferenced. 

Combined, it ends up being almost identical to the union-based `Option` definition above, just without having specified a `union` myself, which is totally possible. However, I prefer to use these transparent building blocks as they do provide some functionality with known behaviors, such as using `ManuallyDrop::drop`, rather than having to call `ptr::drop_in_place(&mut slot.value)`.

I can certainly understand not wanting more `unsafe` in the library, but `MarkedOperation` can only have three mutually exclusive states, and it *drastically* simplifies the code in the operation application and lookahead loops. 

I firmly believe it is fully safe as implemented and used, despite the use of `unsafe` blocks. `unsafe` code is everywhere, even inside `Slice::split_at_mut`. Rust gives us the control to make low-level decisions like this and still build safe interfaces for it.

jonhoo:
Hmm, I can't off the top of my head remember why `Operation` needs to be `Clone`, `Eq`, or `PartialEq`... Does anything break if you remove those bounds?

I don't _think_ exposing an `&` to a shallow-copied value should be problematic. Specifically, the clone should always be equally valid as the original, otherwise we have issues elsewhere (readers read a shallow copy after all!).

As for the `unsafe` stuff, I understand that you're just "unwrapping" the internals to combine the discriminator, I just don't see it as being worth it. For example, `Option` has a number of optimizations for types that are pointers (using `NULL` to indicate `None` for example) that we then wouldn't be getting. We _could_ of course have our own `enum`, but it's not clear to me that that would be any nicer than my proposed two-field struct. As for making the iteration more complicated, I guess I find that surprising looking at the code. From what I see, it _should_ be a pretty simple thing to use mutable references to the log. What issue specifically did you run into? I could give it a try and see what the diff is like?

You are right that these things internally contain `unsafe` that we can "extract", but I don't think doing so without good reason is a good idea. The standard library is pretty well vetted, and we should take advantage of that if we can.

jonhoo:
Specifically, see https://github.com/jonhoo/rust-evmap/compare/collapse..collapse-nounsafe

jonhoo:
@novacrazy Hey! Want to pick this up again and push it over the finish-line? I'm planning another breaking release soon (see the `guards` branch), and this'd be good to include.

novacrazy:
Apologies for the late reply, been very busy lately. I probably won't have time to finish this any time soon, since I'll have to relearn the codebase. 

However, in the past few months I had been working on a very high performance virtual machine with pinhole optimizer, so perhaps I could take what I've learned from that and apply it to this sometime in the next month or two.

jonhoo:
I'm going to go ahead and close this. It's a great idea, but still needs some more work if someone has time to pick it back up. If so, it should go to the new repository over at https://github.com/jonhoo/evmap :)

Issue #16, opened 2019-02-22
ShallowCopy for byte::Bytes


PlaneB:
Seems like byte::Bytes can not be used as type, which is unfortunate, as it should be more or less among the most favored types as it is made for fast data transfer, continuous memory footprint and such.

copy in Bytes is shallow by design, so the following compiler error might be avoidable...

error[E0277]: the trait bound `bytes::Bytes: evmap::ShallowCopy` is not satisfied
   --> src/foo.rs:143:5
    |
143 |     w:WriteHandle<[u8; 32], Bytes, (), BuildNoHashHasher<u32>>,
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `evmap::ShallowCopy` is not implemented for `bytes::Bytes`
    |
    = note: required by `evmap::WriteHandle`

error: aborting due to previous error

jonhoo:
Hmm.. I _think_ it should be okay do have:
```rust
impl ShallowCopy for Bytes {
    unsafe fn shallow_copy(&mut self) -> Self {
        Bytes::from_static(self.as_ref())
    }
}
```
Along those lines, I _also_ think it would be good to add
```rust
impl<'a, T> ShallowCopy for &'a [T] {
    unsafe fn shallow_copy(&mut self) -> Self {
        std::slice::from_raw_parts(self.as_ptr(), self.len())
    }
}
```
I'd be happy to take a look at a PR that adds these :)

PlaneB:
Sorry 'PR' ?

PlaneB:
You want to look into a program that uses those trait implementations? I will try them eventually. What minimal code do you need and why you want to see it?

jonhoo:
I'm not sure I understand what you mean? Those trait implementations would need to be added to `evmap` itself for you to be able to rely on them in an external crate.

PlaneB:
A really? Ok I was under the impression, that the user software implement those traits for Bytes in their code.

jonhoo:
Ah, no, we're limited here by Rust's [_orphan rule_](https://doc.rust-lang.org/book/ch10-02-traits.html#implementing-a-trait-on-a-type):

> One restriction to note with trait implementations is that we can implement a
> trait on a type only if either the trait or the type is local to our crate.
> For example, we can implement standard library traits like `Display` on a
[...]

PR #30, opened 2019-05-28
#16 Bytes shallow copy

kolemannix:
Delivers #16 

- I've made `bytes` an optional dependency. Is this the right way to provide these types of trait `impl`s?
- I've put the conditional `extern crate bytes` at the top of `shallow_copy.rs` for locality, though I'm happy to move it to `lib.rs` for consistency.
- Happy to add tests, if desired, though I did not see tests for the other `ShallowCopy` impls.
- The implementation for a slice reference (commented out) conflicted with the blanket implementation for references: `impl<'a, T> ShallowCopy for &'a T`


src/shallow_copy.rs
```diff
@@ -78,6 +81,20 @@ impl<T> ShallowCopy for Vec<T> {
     }
 }
 
+#[cfg(feature = "bytes")]
+impl ShallowCopy for bytes::Bytes {
+    unsafe fn shallow_copy(&mut self) -> Self {
+        self.clone()
```
kolemannix: I'm new to the `Bytes` crate, but from what I understand, since `Bytes` itself is reference-counted type, I believe a simple `clone()` here satisfies the requirements of an aliasing copy.

An alternative is `Bytes::from(self.as_ref())` but I believe this is doing the same thing.

jonhoo:
Hi! Thanks :tada: I've left some comments in-line :)

src/shallow_copy.rs
```diff
@@ -1,5 +1,8 @@
 //! Types that can be cheaply aliased.
 
+#[cfg(feature = "bytes")]
+extern crate bytes;
```
jonhoo: I think I'd just drop this --- it's no longer needed in the 2018 edition.

src/shallow_copy.rs
jonhoo: Ah, no, that's now quite right. With `ShallowCopy`, only one of the two shallow copies are dropped, the other is simply forgotten. That would cause the given `Bytes` to leak, since the reference count would never hit 0. Take a look at the implementation for `Arc` further up in this file -- it unsafely creates another reference without increasing the count. This is why `ShallowCopy` is so wildly unsafe!

src/shallow_copy.rs
```diff
@@ -78,6 +81,20 @@ impl<T> ShallowCopy for Vec<T> {
...
 
+#[cfg(feature = "bytes")]
+impl ShallowCopy for bytes::Bytes {
+    unsafe fn shallow_copy(&mut self) -> Self {
+        self.clone()
+    }
+}
+
+// This conflicts with the impl below for &'a T
+// impl<'a, T> ShallowCopy for &'a [T] {
```
jonhoo: Is there a particular reason you want this implementation given that the one below already exists?

src/shallow_copy.rs
kolemannix: Updated to use `Bytes::from_static`. I'm wondering if a simple `Bytes::from(self.as_ref())` would work, but I *think* `as_ref` would probably increase the reference count, and that new reference would get passed along to the new `Bytes` instance? So using `from_static` is my attempt to completely disconnect the two `Bytes` instances.

jonhoo: Hmm... I'm not familiar enough with how `Bytes` works internally to say for sure. I think in theory `from_static` should do the right thing though.

jonhoo: I finally had some time today to sit down and read through the implementation of `Bytes`, and I believe `from_static` indeed satisfies the requirements for `ShallowCopy`!

jonhoo:
Thank you!

PR #35, opened 2019-09-11
Update dependencies and features

bstrie:


jonhoo:
This looks great @bstrie, thanks!

@novacrazy I'd be curious if you have an objection to removing `hashbrown` now that it's in `std`?

novacrazy:
The `std` `hashbrown` doesn't use `FxHash` by default, but we weren't using that by default either, though now it's not presented to the user in the `evmap` docs. The `hashbrown` crate itself did provide a better implementation of `FxHash` than the dedicated `fxhash` crate. 

However, this is fine, I think. It may even give a bit better performance to use the `std` version, since `std` is allowed to use nightly features internally on stable IIRC.

jonhoo:
Yup, my thinking here is that people can still depend on `hashbrown` if they want to use their `FxHash`, but us shipping it as a dependency no longer seems necessary.

@bstrie the CI error is https://github.com/xd009642/tarpaulin/issues/190, and is unrelated to your change.

PR #39, opened 2020-01-19
Remove defunct test for non implementation of sync

ptravers:
Hey I was looking through the implementation of this for a separate project and noticed this test which seems to be no longer necessary. I hope this is helpful!

The rust doc issue is now resolved with https://github.com/rust-lang/rust/pull/47833

The documentation in cargo appears to reflect this as Sync is marked !Sync
https://docs.rs/evmap/4.1.1/evmap/struct.ReadHandle.html#impl-Sync

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -431,3 +431,10 @@ where
         .with_meta(meta)
         .construct()
 }
+
+
+#[test]
+fn is_not_sync() {
+    extern crate static_assertions as sa;
+    sa::assert_not_impl_all!(ReadHandle<(), ()>: Send, Sync);
+}
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -431,15 +431,3 @@ where
         .with_meta(meta)
         .construct()
 }
-
-// test that ReadHandle isn't Sync
-// waiting on https://github.com/rust-lang/rust/issues/17606
-//#[test]
-//fn is_not_sync() {
-//    use std::sync;
-//    use std::thread;
-//    let (r, mut w) = new();
-//    w.insert(true, false);
-//    let x = sync::Arc::new(r);
-//    thread::spawn(move || { drop(x); });
-//}
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -164,9 +164,9 @@
 //! }
 //! ```
 //!
-//! `ReadHandle` is not `Sync` as it is not safe to share a single instance
-//! amongst threads. A fresh `ReadHandle` needs to be created for each thread
-//! either by cloning a `ReadHandle` or from a `ReadHandleFactory`.
+//! ReadHandle is not Sync as it is not safe to share a single instance
+//! amongst threads. A fresh ReadHandle needs to be created for each thread
+//! either by cloning a ReadHandle or from a ReadHandleFactory.
 //!
 //! The reason for this is that each ReadHandle assumes that only one
 //! thread operates on it at a time. For details, see the implementation
@@ -184,7 +184,7 @@
 //! is_sync::<ReadHandle<u64, u64>>()
 //!```
 //!
-//! ReadHandle is Send though, since in order to send a ReadHandle, there must
+//! ReadHandle **is** Send though, since in order to send a ReadHandle, there must
 //! be no references to it, so no thread is operating on it.
 //!
 //!```
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -164,13 +164,13 @@
 //! }
 //! ```
 //!
-//! ReadHandle is not Sync as it is not safe to share a single instance
-//! amongst threads. A fresh ReadHandle needs to be created for each thread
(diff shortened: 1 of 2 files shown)


jonhoo:
Ah, the intention behind having a test was that I don't always remember to check the documentation with every change that I make. I _do_ run the test suite each time though, so if it could catch `ReadHandle` accidentally becoming `Sync`, that'd be much better. The test seems to currently be commented out though, and I forget exactly why. Maybe you could look into a way to bring the test back? That'd be super helpful!

ptravers:
Hey sorry this totally slipped by me in a slew of github comments. Sure happy to look at that. Right now it won't compile. Happy to take a look at writing an assertion that it will fail to compile due to the lack of Send or investigating a different solution for checking for the lack of Send. I think right now the compiler spews as RefCell cannot be passed to a new thread without Send implemented? But I will check a little later today!

jonhoo:
No worries, life gets busy some times!

I _think_ `ReadHandle` should be `Send`, but not `Sync`.

src/lib.rs
```diff
@@ -432,14 +432,9 @@ where
...
-//    let (r, mut w) = new();
-//    w.insert(true, false);
-//    let x = sync::Arc::new(r);
-//    thread::spawn(move || { drop(x); });
-//}
+
+#[test]
+fn is_not_sync() {
+    extern crate static_assertions as sa;
+    sa::assert_not_impl_all!(ReadHandle<(), ()>: Send, Sync);
```
ptravers: Here's the docs for the macro

https://docs.rs/static_assertions/1.1.0/static_assertions/macro.assert_not_impl_all.html

ptravers:
hey I think this is now doing what we discussed!

src/lib.rs
jonhoo: Hmm, this doesn't seem quite right. `ReadHandle` _should_, I believe, be `Send`, it just should _not_ be `Sync`.

ptravers: Yeah so I found the documentation a little confusing. But I believe this is correct given the following:

```
Although u32 implements From<u16>, it does not implement Into<usize>:

assert_not_impl_all!(u32: From<u16>, Into<usize>);
```

I will take a look at breaking the test and get back to you tomorrow.

per: https://docs.rs/static_assertions/1.1.0/static_assertions/macro.assert_not_impl_all.html

jonhoo: Reading that, I _think_ the meaning is to check that it does not implement _all_ of the passed traits. So in the above, the test is "`u32` does not implement _both_ `From<u16>` _and_ `Into<usize>`". I think I would just write two assertions. One that the type _is_ `Send`, and on that the type is _not_ `Sync`.

src/lib.rs
```diff
@@ -164,6 +164,26 @@
...
+//!```
+//! use evmap::ReadHandle;
+//!
+//! fn is_send<T: Send>() {
+//!   // dummy function just used for its parameterized type bound
+//! }
+//!
+//! is_send::<ReadHandle<u64, u64>>()
+//!```
+//!
```
ptravers: I refactored to put this in a doc test. My only concern is that it doesn't test exclusively that the trait is inherited so can pass if any line of code doesn't compile.

I refactored to this because it seemed that:
a) It was better supported in the rust ecosystem since the language itself uses this doctest macro
b) I couldn't get the static assertions macro to fail in ways that I would expect and I did not fully grok how it worked. Also admittedly could make more effort on that front.

jonhoo: Yup, this looks like a good approach!

src/lib.rs
```diff
@@ -164,6 +164,36 @@
 //! }
 //! ```
 //!
+//! `ReadHandle` is not `Sync` as it is not safe to share a single instance
+//! amongst threads. A fresh `ReadHandle` needs to be created for each thread
+//! either by cloning a `ReadHandle` or from a `ReadHandleFactory`.
```
jonhoo: I would add to this:
> The reason for this is that each `ReadHandle` assumes that only one thread operates on it at a time. For details, see the implementation comments on `ReadHandle`. `ReadHandle` _is_ `Send` though, since in order to send a `ReadHandle`, there must be no references to it, so no thread is operating on it.

ptravers: done!

ptravers: words are the same but moved them about a wee bit

jonhoo: It would be good to also leave the backtics for code blocks and the **emphasis** on **is** `Send`.

ptravers: done

jonhoo: Ah, sorry, I think phrased myself poorly when it came to the backtics. I mean to leave the backtics **in** as well as the emphasis :p

ptravers: ha no worries! I think we're there now?

jonhoo:
Beautiful, thank you!

PR #40, opened 2020-01-29
Bump bytes dependency to 0.5

lukesteensen:
This seems to work as-is and helps interop with other crates in the ecosystem that have upgraded.

Thanks!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,7 @@ maintenance = { status = "passively-maintained" }
 [dependencies]
 indexmap = { version = "1.1.0", optional = true }
 smallvec = { version = "1.0.0", optional = true }
-bytes = { version = "0.4.12", optional = true }
+bytes = { version = "0.5.4", optional = true }
 
 [features]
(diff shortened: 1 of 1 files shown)


jonhoo:
Ah, interesting. So, this is technically a backwards-incompatible change:
https://github.com/jonhoo/rust-evmap/blob/5280dd1189078a54377bf3375ac3fa5b442fd943/src/shallow_copy.rs#L82
But I think we should probably adopt it anyway. I'll merge and release.

lukesteensen:
Yeah, I think it will require a version bump unfortunately. That impl is actually how I came across it.

jonhoo:
Could you rebase/merge with `master` so we get working CI?

jonhoo:
Closed in d56d013d65858eb85d2929fe6b8398ef2919e490 and released as 8.0.0 :tada:

lukesteensen:
Sorry, looks like I was too slow :laughing:

Thanks @jonhoo!

PR #44, opened 2020-02-07
hashbag for large bags

jonhoo:
This avoids a huge performance collapse when there are many values for a
given key and the user wants to remove one. We do this by switching
large values over to a `hashbag` for backing storage. This change means
that users no longer get a `&[V]`, but instead a `&Values<V>`, and also
that we require `V: Hash + Eq`.

With this change, I also made `smallvec` non-optional.

I also replaced the awkward `mem::forget` code with `ManuallyDrop` which
seems like a much safer approach. The failure mode is now memory leaks
rather than double-frees, which seems like a huge win.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,9 +24,10 @@ maintenance = { status = "passively-maintained" }
 
 [dependencies]
 indexmap = { version = "1.1.0", optional = true }
-smallvec = { version = "1.0.0", optional = true }
+smallvec = "1.0.0"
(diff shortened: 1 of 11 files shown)


jonhoo:
Checked manually that `miri` passes on `rustc 1.43.0-nightly (58b834344 2020-02-05)`.

jonhoo:
`miri` still passes on a slightly older nightly where it is present.

Issue #47, opened 2020-03-17
Blanket implementation of ShallowCopy for Copy types


MayaWolf:
Hi!

First off, thank you for this great library! It's perfect for the application I'm developing.
One thing that's been making it a little awkward in use is the requirement for `ShallowCopy`. As far as I can see, it's really intended to be a bit of an extended copy - anything that _can_ be copied should be, with special cases for certain types that could normally only implement `Clone` - correct?
If so, I think the library could benefit (both in use and in internal code) from a blanket implementation for Copy, like so:
```rs
impl<T> ShallowCopy for T where T: Copy {
    unsafe fn shallow_copy(&self) -> ManuallyDrop<Self> {
        ManuallyDrop::new(*self)
    }
}
```

What do you think?

jonhoo:
Hi there!

Glad you've found it useful :) There is actually already such an impl, it's just not a blanket implementation because those can be painful for downstream crates to deal with. Instead, you opt in by using the [`CopyValue` wrapper type](https://docs.rs/evmap/9/evmap/shallow_copy/struct.CopyValue.html). How's that?

MayaWolf:
Thanks for your response!

I've been looking into it some more (full disclosure, I'm both working with partially generated code and trying to write a derive macro for ShallowCopy, so mine might be a bit of a special case) and I think the main issue I'm running into is actually that there's no ShallowCopy impl in place for `Option<ShallowCopy>` - is there a conscious reason this doesn't exist, or do you think such an impl could be added?

jonhoo:
Interesting, that does sound like a niche use-case indeed, but I'm still keen to support it!

Yes, indeed, there's a missing `impl` for `Option<T> where T: ShallowCopy`. I'd be happy to accept a PR! I think _technically_ it's a breaking change, but given that I _just_ released `9.0.0` we should be okay to yank it.

MayaWolf:
Awesome! I'm going to be busy with the refactor of my codebase to use evmap for a little while longer, but once I'm done and I've gotten a chance to test the changes, I'm going to submit a PR!

PR #48, opened 2020-03-25
WIP: ShallowCopy for Option<T> (and more)

MayaWolf:
Hi there!

As mentioned in #47, I'm using evmap in a (soon-to-be) production project, and ended up making some changes to make evmap easier for _my personal_ use - and I figured some of them might be interesting to you as well. I am merely presenting this to you as a list of things to pick and choose from - tell me which ones you'd be interested in merging, and I'll remove the rest from this PR. Also if there are any code style/documentation/naming/other changes you'd like me to make (including splitting things up into multiple commits/PRs), just let me know!

- Added an evmap-derive proc-macro crate for ShallowCopy
- Removed the requirement for Send + Sync on `Predicate` for `reject` - it's not actually sent across threads, so there's no need for it from what I could see
- Added an IntoIterator implementation for ReadGuard (allows doing `map.get(&key).iter().flatten()`)
- Made read() return an Option instead, which allows many of the MapReadRef methods to return non-wrapped values (the way I understood it it can't be modified or destroyed while the lock is held anyway)
- Added `single` functions to the read modules - allows for simple and efficient reading of single values
- Added `contains_value` convenience function to the read modules
- Added ShallowCopy implementation for Option<T>
- Added Default derive for ShallowCopy
- Added defaults for MapReadRef type parameters M and S to make it easier in use

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/rust-evmap/48)

### DIFF
--- a/evmap-derive/Cargo.toml
+++ b/evmap-derive/Cargo.toml
@@ -1,0 +1,30 @@
+[package]
+name = "evmap-derive"
+version = "9.0.0"
+edition = "2018"
+
+description = "A lock-free, eventually consistent, concurrent multi-value map."
+readme = "README.md"
+
+authors = ["Jon Gjengset <jon@thesquareplanet.com>"]
+
+documentation = "https://docs.rs/evmap"
+homepage = "https://github.com/jonhoo/rust-evmap"
+repository = "https://github.com/jonhoo/rust-evmap.git"
+
+keywords = ["map","multi-value","lock-free"]
+categories = ["concurrency", "data-structures"]
+
+license = "MIT/Apache-2.0"
+
+[lib]
+proc-macro = true
+
+[dependencies]
+syn = "1.0.16"
+quote = "1.0.2"
+proc-macro2 = "1.0.9"
+
+[dev-dependencies]
+evmap = { path = "../", version = "9.0.0" }
+trybuild = "1.0.24"
--- a/evmap-derive/src/lib.rs
+++ b/evmap-derive/src/lib.rs
@@ -1,0 +1,174 @@
+//! This crate provides procedural derive macros to simplify the usage of `evmap`.
+//!
+//! Currently, only `#[derive(ShallowCopy)]` is supported; see below.
+#![deny(missing_docs)]
+use proc_macro2::{Ident, TokenStream};
+use quote::{format_ident, quote, quote_spanned};
+use syn::spanned::Spanned;
+use syn::{
+    parse_macro_input, parse_quote, Data, DeriveInput, Fields, GenericParam, Generics, Index,
+};
+
+/// Implementation for `#[derive(ShallowCopy)]`
+///
+/// evmap provides the [`ShallowCopy`](evmap::shallow_copy::ShallowCopy) trait, which allows you to
+/// cheaply alias types that don't otherwise implement `Copy`. Basic implementations are provided
+/// for common types such as `String` and `Vec`, but it must be implemented manually for structs
+/// using these types.
+///
+/// This macro attempts to simplify this task. It only works on types whose members all implement
+/// `ShallowCopy`. If this is not possible, consider using
+/// [`CopyValue`](evmap::shallow_copy::CopyValue), `Box`, or `Arc` instead.
+///
+/// ## Usage example
+/// ```
+/// # use evmap_derive::ShallowCopy;
+/// #[derive(ShallowCopy)]
+/// struct Thing { field: i32 }
+///
+/// #[derive(ShallowCopy)]
+/// struct Generic<T> { field: T }
+///
+/// #[derive(ShallowCopy)]
+/// enum Things<T> { One(Thing), Two(Generic<T>) }
+/// ```
+///
+/// ## Generated implementations
+/// The generated implementation calls
+/// [`shallow_copy`](evmap::shallow_copy::ShallowCopy::shallow_copy) on all the members of the
+/// type, and lifts the `ManuallyDrop` wrappers to the top-level return type.
+///
+/// For generic types, the derive adds `ShallowCopy` bounds to all the type parameters.
+///
+/// For instance, for the following code...
+/// ```
+/// # use evmap_derive::ShallowCopy;
+/// #[derive(ShallowCopy)]
+/// struct Generic<T> { field: T }
+/// ```
+/// ...the derive generates...
+/// ```
+/// # use evmap::shallow_copy::ShallowCopy;
+/// # use std::mem::ManuallyDrop;
+/// # struct Generic<T> { field: T }
+/// impl<T: ShallowCopy> ShallowCopy for Generic<T> {
+///     unsafe fn shallow_copy(&self) -> ManuallyDrop<Self> {
+///         ManuallyDrop::new(Self {
+///             field: ManuallyDrop::into_inner(ShallowCopy::shallow_copy(&self.field))
+///         })
+///     }
+/// }
+/// ```
+#[proc_macro_derive(ShallowCopy)]
+pub fn derive_shallow_copy(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
+    let input = parse_macro_input!(input as DeriveInput);
+    let name = input.ident;
+    let generics = add_trait_bounds(input.generics);
+    let (impl_generics, ty_generics, where_clause) = generics.split_for_impl();
+    let copy = fields(&input.data, &name);
+    proc_macro::TokenStream::from(quote! {
+        impl #impl_generics evmap::shallow_copy::ShallowCopy for #name #ty_generics #where_clause {
+            unsafe fn shallow_copy(&self) -> std::mem::ManuallyDrop<Self> {
+                #copy
+            }
+        }
+    })
+}
+
+fn add_trait_bounds(mut generics: Generics) -> Generics {
+    for param in &mut generics.params {
+        if let GenericParam::Type(ref mut type_param) = *param {
+            type_param
+                .bounds
+                .push(parse_quote!(evmap::shallow_copy::ShallowCopy));
+        }
+    }
+    generics
+}
+
+fn fields(data: &Data, type_name: &Ident) -> TokenStream {
+    match data {
+        Data::Struct(data) => match &data.fields {
+            Fields::Named(fields) => {
+                let recurse = fields.named.iter().map(|f| {
+                        let name = &f.ident;
+                        quote_spanned! {f.span()=>
+                            #name: std::mem::ManuallyDrop::into_inner(
+                                evmap::shallow_copy::ShallowCopy::shallow_copy(&self.#name)
+                            )
+                        }
+                    });
+                quote! {
+                    std::mem::ManuallyDrop::new(Self { #(#recurse,)* })
+                }
+            }
+            Fields::Unnamed(fields) => {
+                let recurse = fields.unnamed.iter().enumerate().map(|(i, f)| {
+                        let index = Index::from(i);
+                        quote_spanned! {f.span()=>
+                            std::mem::ManuallyDrop::into_inner(
+                                evmap::shallow_copy::ShallowCopy::shallow_copy(&self.#index)
+                            )
+                        }
+                    });
+                quote! {
+                    std::mem::ManuallyDrop::new(#type_name(#(#recurse,)*))
+                }
+            }
+            Fields::Unit => quote!(std::mem::ManuallyDrop::new(#type_name)),
+        },
+        Data::Enum(data) => {
+            let recurse = data.variants.iter().map(|f| {
+                let (names, fields) = match &f.fields {
+                    Fields::Named(fields) => {
+                        let field_names = f.fields.iter().map(|field| {
+                            let ident = field.ident.as_ref().unwrap();
+                            quote_spanned! {
+                                field.span()=> #ident
+                            }
+                        });
+                        let recurse = fields.named.iter().map(|f| {
+                            let name = f.ident.as_ref().unwrap();
+                            quote_spanned! {f.span()=>
+                                #name: std::mem::ManuallyDrop::into_inner(
+                                    evmap::shallow_copy::ShallowCopy::shallow_copy(#name)
+                                )
+                            }
+                        });
+                        (quote! { {#(#field_names,)*} }, quote! { { #(#recurse,)* } })
+                    }
+                    Fields::Unnamed(fields) => {
+                        let field_names = f.fields.iter().enumerate().map(|(i, field)| {
+                            let ident = format_ident!("x{}", i);
+                            quote_spanned! {
+                                field.span()=> #ident
+                            }
+                        });
+                        let recurse = fields.unnamed.iter().enumerate().map(|(i, f)| {
+                            let ident = format_ident!("x{}", i);
+                            quote_spanned! {f.span()=>
+                                std::mem::ManuallyDrop::into_inner(
+                                    evmap::shallow_copy::ShallowCopy::shallow_copy(#ident)
+                                )
+                            }
+                        });
+                        (quote! { (#(#field_names,)*) }, quote! { (#(#recurse,)*) })
+                    }
+                    Fields::Unit => {
+                        (quote!(), quote!())
+                    }
+                };
+                let name = &f.ident;
+                quote_spanned! {f.span()=>
+                    #type_name::#name#names => std::mem::ManuallyDrop::new(#type_name::#name#fields)
+                }
+            });
+            quote! {
+                match self {
+                    #(#recurse,)*
+                }
+            }
+        }
+        Data::Union(_) => unimplemented!("Unions are not supported"),
+    }
+}
--- a/evmap-derive/tests/failing/lib.rs
+++ b/evmap-derive/tests/failing/lib.rs
@@ -1,0 +1,25 @@
+use evmap_derive::ShallowCopy;
+use evmap::shallow_copy::ShallowCopy as _;
+
+#[derive(ShallowCopy)]
+struct Broken {
+	f1: i32,
+	f2: std::cell::Cell<()>,
+}
+
+#[derive(ShallowCopy)]
+struct AlsoBroken(i32, std::cell::Cell<()>);
+
+#[derive(ShallowCopy)]
+struct Generic<T> {
+	f1: T
+}
+
+fn main() {
+	let broken = Broken { f1: 0, f2: std::cell::Cell::new(()) };
+	broken.shallow_copy();
+	let also_broken = AlsoBroken(0, std::cell::Cell::new(()));
+	broken.shallow_copy();
+	let generic = Generic { f1: std::cell::Cell::new(()) };
+	generic.shallow_copy();
+}
--- a/evmap-derive/tests/lib.rs
+++ b/evmap-derive/tests/lib.rs
@@ -1,0 +1,65 @@
+use evmap::shallow_copy::ShallowCopy as _;
+use evmap_derive::ShallowCopy;
+use std::sync::Arc;
+
+#[derive(ShallowCopy)]
+enum Message {
+    Quit,
+    ChangeColor(i32, i32, i32),
+    Move { x: i32, y: i32 },
+    Write(String),
+}
+
+#[derive(ShallowCopy)]
+struct Shallow;
+
+#[derive(ShallowCopy)]
+struct Tuple(i32, String, Shallow);
+
+#[derive(ShallowCopy)]
+struct Test {
+    f1: i32,
+    f2: (i32, i32),
+    f3: String,
+    f4: Arc<String>,
+    f5: Shallow,
+    f6: evmap::shallow_copy::CopyValue<i32>,
+    f7: Tuple,
+}
+
+#[derive(ShallowCopy)]
+struct Generic<T> {
+    f1: T,
+}
+
+#[test]
+fn test() {
+    unsafe {
+        let message = Message::Quit;
+        message.shallow_copy();
+        let shallow = Shallow;
+        shallow.shallow_copy();
+        let tuple = Tuple(0, "test".to_owned(), Shallow);
+        tuple.shallow_copy();
+        let test = Test {
+            f1: 0,
+            f2: (1, 2),
+            f3: "test".to_owned(),
+            f4: Arc::new("test".to_owned()),
+            f5: shallow,
+            f6: 3.into(),
+            f7: tuple,
+        };
+        test.shallow_copy();
+        let generic = Generic {
+            f1: "test".to_owned(),
+        };
+        generic.shallow_copy();
+    }
+}
+
+#[test]
+fn failing() {
+    let t = trybuild::TestCases::new();
+    t.compile_fail("tests/failing/lib.rs");
+}
--- a/src/read/mod.rs
+++ b/src/read/mod.rs
@@ -198,21 +198,25 @@ where
     ///
     /// While the reference lives, the map cannot be refreshed.
     ///
+    /// If no refresh has happened, or the map has been destroyed, this function returns `None`.
+    ///
     /// See [`MapReadRef`].
-    pub fn read(&self) -> MapReadRef<'_, K, V, M, S> {
-        MapReadRef {
-            guard: self.handle(),
+    pub fn read(&self) -> Option<MapReadRef<'_, K, V, M, S>> {
+        let guard = self.handle()?;
+        if !guard.is_ready() {
+            return None;
         }
+        Some(MapReadRef { guard })
     }
 
     /// Returns the number of non-empty keys present in the map.
     pub fn len(&self) -> usize {
-        self.read().len()
+        self.read().map_or(0, |x| x.len())
     }
 
     /// Returns true if the map contains no elements.
     pub fn is_empty(&self) -> bool {
-        self.read().is_empty()
+        self.read().map_or(true, |x| x.is_empty())
     }
 
     /// Get the current meta value.
@@ -286,7 +290,7 @@ where
     ///
     /// See [`WriteHandle::destroy`].
     pub fn is_destroyed(&self) -> bool {
-        self.read().is_destroyed()
+        self.handle().is_none()
     }
 
     /// Returns true if the map contains any values for the specified key.
@@ -298,7 +302,7 @@ where
         K: Borrow<Q>,
         Q: Hash + Eq,
     {
-        self.read().contains_key(key)
+        self.read().map_or(false, |x| x.contains_key(key))
     }
 
     /// Read all values in the map, and transform them into a new collection.
@@ -307,7 +311,7 @@ where
         Map: FnMut(&K, &Values<V, S>) -> Target,
         Collector: FromIterator<Target>,
     {
-        Collector::from_iter(self.read().iter().map(|(k, v)| f(k, v)))
+        Collector::from_iter(self.read().iter().flatten().map(|(k, v)| f(k, v)))
     }
 }
 
--- a/src/read/mod.rs
+++ b/src/read/mod.rs
@@ -120,6 +120,7 @@ where
 impl<K, V, M, S> ReadHandle<K, V, M, S>
 where
     K: Eq + Hash,
+    V: Eq + Hash,
     S: BuildHasher,
     M: Clone,
 {
@@ -257,6 +258,30 @@ where
         Some(self.get_raw(key.borrow())?.map_ref(Values::user_friendly))
     }
 
+    /// Returns a guarded reference to _one_ value corresponding to the key.
+    ///
+    /// This is mostly intended for use when you are working with no more than one value per key.
+    /// If there are multiple values stored for this key, there are no guarantees to which element
+    /// is returned.
+    ///
+    /// While the guard lives, the map cannot be refreshed.
+    ///
+    /// The key may be any borrowed form of the map's key type, but `Hash` and `Eq` on the borrowed
+    /// form must match those for the key type.
+    ///
+    /// Note that not all writes will be included with this read -- only those that have been
+    /// refreshed by the writer. If no refresh has happened, or the map has been destroyed, this
+    /// function returns `None`.
+    #[inline]
+    pub fn get_one<'rh, Q: ?Sized>(&'rh self, key: &'_ Q) -> Option<ReadGuard<'rh, V>>
+    where
+        K: Borrow<Q>,
+        Q: Hash + Eq,
+    {
+        self.get_raw(key.borrow())?
+            .map_opt(|x| x.user_friendly().get_one())
+    }
+
     /// Returns a guarded reference to the values corresponding to the key along with the map
     /// meta.
     ///
@@ -303,6 +328,22 @@ where
         Q: Hash + Eq,
     {
         self.read().map_or(false, |x| x.contains_key(key))
+    }
+
+    /// Returns true if the map contains the specified value for the specified key.
+    ///
+    /// The key and value may be any borrowed form of the map's respective types, but `Hash` and
+    /// `Eq` on the borrowed form *must* match.
+    pub fn contains_value<Q: ?Sized, W: ?Sized>(&self, key: &Q, value: &W) -> bool
+    where
+        K: Borrow<Q>,
+        V: Borrow<W>,
+        Q: Hash + Eq,
+        W: Hash + Eq,
+    {
+        self.get_raw(key.borrow())
+            .map(|x| x.user_friendly().contains(value))
+            .unwrap_or(false)
     }
 
     /// Read all values in the map, and transform them into a new collection.
--- a/src/read/read_ref.rs
+++ b/src/read/read_ref.rs
@@ -17,6 +17,7 @@ use std::mem::ManuallyDrop;
 pub struct MapReadRef<'rh, K, V, M = (), S = RandomState>
 where
     K: Hash + Eq,
+    V: Eq + Hash,
     S: BuildHasher,
 {
     pub(super) guard: ReadGuard<'rh, Inner<K, ManuallyDrop<V>, M, S>>,
@@ -25,6 +26,7 @@ where
 impl<'rh, K, V, M, S> MapReadRef<'rh, K, V, M, S>
 where
     K: Hash + Eq,
+    V: Eq + Hash,
     S: BuildHasher,
 {
     /// Iterate over all key + valuesets in the map.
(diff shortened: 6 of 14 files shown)


jonhoo:
Oooh, exciting! I'm a bit strapped for time, but here are some quick thoughts to get the conversation going:

---
> Added an evmap-derive proc-macro crate for `ShallowCopy`

How often does this come up? My expectation was that implementations of `ShallowCopy` should be _very_ rare, and that users should be extremely careful with them. I'm hesitant to hide it from the user unless this is a very common occurrence.

---

> Removed the requirement for `Send + Sync` on `Predicate` for reject - it's not actually sent across threads, so there's no need for it from what I could see.

Removing `Sync` is probably fine, but removing `Send` means that `WriteHandle` is no longer `Send`, which would be sad. (`WriteHandle` is already not `Sync` since it contains a `ReadHandle`)

---

> Added an `IntoIterator` implementation for `ReadGuard` (allows doing `map.get(&key).iter().flatten()`)

Seems like a great idea!

---

> Made `read()` return an `Option` instead, which allows many of the `MapReadRef` methods to return non-wrapped values (the way I understood it it can't be modified or destroyed while the lock is held anyway)

This seems reasonable to me. It's a breaking change, but a good one. The docs for `read()` should be updated to reflect what `None` means though.

---

> Added `single` functions to the read modules - allows for simple and efficient reading of single values

I like this idea. What do you think of the name `get_first`? Also, docs still refer to multiple values, but I'm guessing you've just left those for later.

---

> Added `contains_value` convenience function to the `read` modules

Makes sense to me!

---

> Added `ShallowCopy` implementation for `Option`

Absolutely.

---

> Added `Default` derive for `ShallowCopy`

Do you mean `CopyValue`? Seems fine to me.

---

> Added defaults for `MapReadRef` type parameters `M` and `S` to make it easier in use

Yeah, that makes sense.

---

I'll also leave some notes about specifics that you didn't list above! As for commits, having distinct changes in their own commit is _nice_, but not _necessary_. It mostly makes pinpointing where bugs are introduced years from now easier :)

src/read/mod.rs
```diff
@@ -120,6 +120,7 @@ where
 impl<K, V, M, S> ReadHandle<K, V, M, S>
 where
     K: Eq + Hash,
+    V: Eq + Hash,
```
jonhoo: I don't see why these have been added everywhere? That seems like an unfortunate requirement to add.

src/read/mod.rs
```diff
@@ -307,7 +341,7 @@ where
         Map: FnMut(&K, &Values<V, S>) -> Target,
         Collector: FromIterator<Target>,
     {
-        Collector::from_iter(self.read().iter().map(|(k, v)| f(k, v)))
+        Collector::from_iter(self.read().iter().flatten().map(|(k, v)| f(k, v)))
```
jonhoo: What purpose does this change serve?

src/read/read_ref.rs
```diff
@@ -65,23 +62,28 @@ where
     /// Note that not all writes will be included with this read -- only those that have been
     /// refreshed by the writer. If no refresh has happened, or the map has been destroyed, this
     /// function returns `None`.
-    pub fn get<'a, Q: ?Sized>(&'a self, key: &'_ Q) -> Option<&'a Values<V, S>>
+    pub fn get<Q: ?Sized>(&self, key: &Q) -> Option<&Values<V, S>>
```
jonhoo: This change is not okay, because it means the return value is now also tied to the lifetime of the provided key.

src/read/read_ref.rs
```diff
@@ -65,23 +62,28 @@ where
...
+    /// The key may be any borrowed form of the map's key type, but `Hash` and `Eq` on the borrowed
+    /// form *must* match those for the key type.
     ///
-    /// See [`WriteHandle::destroy`].
-    pub fn is_destroyed(&self) -> bool {
-        self.guard.is_none()
+    /// Note that not all writes will be included with this read -- only those that have been
+    /// refreshed by the writer. If no refresh has happened, or the map has been destroyed, this
+    /// function returns `None`.
+    pub fn single<Q: ?Sized>(&'rh self, key: &'rh Q) -> Option<&'rh V>
```
jonhoo: The lifetime on `key` should be `'_` to avoid making the return type tied to the the key's lifetime.

src/values.rs
```diff
@@ -70,6 +70,16 @@ impl<T, S> Values<T, S> {
             ValuesInner::Long(ref v) => ValuesIter::Long(v.iter()),
         }
     }
+    
+    /// An iterator visiting all elements in arbitrary order.
```
jonhoo: The docs here seem... wrong :p

src/read/mod.rs
MayaWolf: They're already requirements of the map itself - so unless I'm missing anything it only makes sense for them to be there as well, no? I think it was necessary for adding the `contains_value` helpers.

src/read/mod.rs
MayaWolf: `read()` now returns an Option - this gets rid of it.

src/read/mod.rs
jonhoo: Ah, I see, they are requirements on `WriteHandle` because we unconditionally support `retain` :'( All right, I guess it makes sense to have them here too then.

src/read/mod.rs
jonhoo: Ah, right, because `Option::iter` iterates over the one element in the `Option`. That's neat I suppose :p

src/read/mod.rs
MayaWolf: I'm pretty sure values would have to be `Eq + Hash` anyway due to the use in `hashbag`, no?

MayaWolf:
I'll fix the other things you brought up in a moment here, but quick responses to your comments:
>How often does this come up? My expectation was that implementations of ShallowCopy should be very rare, and that users should be extremely careful with them. I'm hesitant to hide it from the user unless this is a very common occurrence.
It's extremely useful when you have structs as values that don't exclusively contain `Copy` values (`String` being the biggest offender). If you can see a better way of doing that, please let me know, but that's basically what I was using it for.

>Do you mean CopyValue? Seems fine to me.
Yeah, that's what I meant.

MayaWolf:
I've added the requested changes in an extra commit for quick review - once you're happy with all of it, I'll rebase it into something a little more structured.

Added clarification to the derive macro point: I'm using multiple structs of primitives and strings, so they can't be `Copy` and I'd therefore have to manually write `ShallowCopy` impls for each of them to use them as values in evmap.

src/values.rs
```diff
@@ -71,10 +71,8 @@ impl<T, S> Values<T, S> {
         }
     }
     
-    /// An iterator visiting all elements in arbitrary order.
-    ///
-    /// The iterator element type is &'a T.
-    pub fn single(&self) -> Option<&T> {
+    /// Returns the first element.
```
jonhoo: I think we may want to expand on this a little. For example, what does "first" mean if the backing store is a `HashBag`?

evmap-derive/src/lib.rs
```diff
@@ -0,0 +1,77 @@
...
+                Fields::Unit => {
+                    quote!(())
+                }
+            }
+        }
+        Data::Enum(data) => {
+            let recurse = data.variants.iter().map(|f| {
+                let name = &f.ident;
+                quote_spanned! {f.span()=>
+                    #type_name::#name(x) => std::mem::ManuallyDrop::new(#type_name::#name(std::mem::ManuallyDrop::into_inner(evmap::shallow_copy::ShallowCopy::shallow_copy(x))))
```
jonhoo: I think this assumes that all enum variants consist only of a single tuple-style field?

evmap-derive/src/lib.rs
```diff
@@ -0,0 +1,77 @@
+use proc_macro2::{Ident, TokenStream};
+use quote::{quote, quote_spanned};
+use syn::spanned::Spanned;
+use syn::{parse_macro_input, parse_quote, Data, DeriveInput, Fields, GenericParam, Generics, Index};
+
+#[proc_macro_derive(ShallowCopy)]
+pub fn derive_heap_size(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
```
jonhoo: This name seems wrong?

MayaWolf: Lol yes you can tell I based it on the example from `syn`.

jonhoo:
> If you can see a better way of doing that, please let me know, but that's basically what I was using it for.

So, in my use-cases, I "solve" this by allocating the offending value type behind a `Box` or `Arc`. But your approach is certainly less wasteful. I worry about people using this without realizing quite what they are opting into though. `ShallowCopy` is deeply unsafe, and users should _really_ know what's going on to use it. Then again, maybe a derive is a step in the right direction there, because there isn't a way for them to really mess it up (same argument as `#[pin_project]` in a way).

I'm inclined to include `evmap-derive` as well. We'll want to make sure it's well document exactly what it does though, and probably also include some examples + tests.

MayaWolf:
>maybe a derive is a step in the right direction there, because there isn't a way for them to really mess it up

That was my thought process as well - it only works on structs/enums if all of their fields already implement `ShallowCopy` - you can't really apply it to types that don't technically already support it.

Speaking of which, I wonder whether there's a way around having to unwrap and then re-wrap with `ManuallyDrop` in the macro - any ideas? Would it be possible for the `ShallowCopy` impls to not have to return `ManuallyDrop` directly (as it was before 9.0.0), but to create this wrapper in `evmap`'s code instead?

jonhoo:
Since `ManuallyDrop` is `#[repr(transparent)]` I think technically you can just cast without having to do the whole unwrap-rewrap dance. I want to keep the signature the way it is to make the API as misuse-resistant as possible, even within `evmap`. If it actually returned `Self`, then you can bet at _some_ point, _someone_ (quite possibly me) will forget that it needs to be wrapped and carry it along and accidentally drop it.

I think I agree with you re `evmap-derive`. Let's keep it.

MayaWolf:
I made a few more changes - let me know whether the name `get_one` works for you or whether we should come up with something better, still (I figured if "first" could be confusing in the doc comment, it could be doubly so in the name).

One more thing: Do you think `CopyValue` could/should be working on `Clone` types rather than only `Copy` ones? I don't need this personally (due to the derive macro), but I also can't really see any reason why it shouldn't.

evmap-derive/src/lib.rs
```diff
@@ -61,9 +61,21 @@ fn fields(data: &Data, type_name: &Ident) -> TokenStream {
...
+                let fields = f.fields.iter().enumerate().map(|(i, field)| {
+                    let ident = format_ident!("x{}", i);
+                    quote_spanned! {field.span()=>
+                       std::mem::ManuallyDrop::into_inner(evmap::shallow_copy::ShallowCopy::shallow_copy(#ident))
+                    }
+                });
                 let name = &f.ident;
                 quote_spanned! {f.span()=>
-                    #type_name::#name(x) => std::mem::ManuallyDrop::new(#type_name::#name(std::mem::ManuallyDrop::into_inner(evmap::shallow_copy::ShallowCopy::shallow_copy(x))))
+                    #type_name::#name(#(#field_names,)*) => std::mem::ManuallyDrop::new(#type_name::#name(#(#fields,)*))
```
jonhoo: This still seems like it would only support tuple-style enum variants? Having some tests here would be very good.

src/read/mod.rs
```diff
@@ -254,7 +254,9 @@ where
         Some(self.get_raw(key.borrow())?.map_ref(Values::user_friendly))
     }
 
-    /// Returns a guarded reference to the first value corresponding to the key.
+    /// Returns a guarded reference to _one_ value corresponding to the key.
+    /// This is mostly intended for use when you are working with no more than one value per key.
```
jonhoo: Super nitpicky, but could you leave an empty line after the first sentence? Makes it a little easier to visually parse out the "important" parts of the docs.

jonhoo: And same for the other `get_one`.

jonhoo:
So, my thinking for `get_first` was that it gets the "first item it sees", no matter which one it is. The implication being that it is efficient even when there are potentially many items. I'm okay with `get_one` too if you prefer that. Alternatively, `get_any` might work, although maybe `any` is confusing in this context?

And no, sadly, `CopyValue` _has_ to be only explicitly for `Copy` values. If it could take `Clone` values, you would end up with multiple clones of the value, **both of which need to be freed**. `ShallowCopy` is specifically used in a setting where only _one_ of the copies is freed, and so allowing `T: Clone` would lead to a continuous memory leak. This is the kind of stuff that makes me really paranoid about anything that makes `ShallowCopy` "more accessible" :sweat_smile:

MayaWolf:
I added some tests and fixed the existing ones - let me know if there's anything else!

MayaWolf:
I added one more convenience function to cover a pattern I found myself using very commonly, `get_owned`. It returns a cloned value if it exists, and the default value otherwise, and it saves you from writing `.cloned().unwrap_or_default()` all the time when doing upserts.

Again, I'm not sure how specific that is to my use case, so I'll leave it up to you whether it's something you'd be interested in merging in.

evmap-derive/Cargo.toml
```diff
@@ -26,4 +26,4 @@ quote = "1.0.2"
 proc-macro2 = "1.0.9"
 
 [dev-dependencies]
-evmap = "9.0.0"
+evmap = { path = "../" }
```
jonhoo: I think the general advice here is to give _both_ a version _and_ a path. That way `git` builds and patched builds continue to work.

evmap-derive/tests/lib.rs
```diff
@@ -0,0 +1,28 @@
...
+
+#[derive(ShallowCopy)]
+enum Message {
+	Quit,
+	ChangeColor(i32, i32, i32),
+	Move { x: i32, y: i32 },
+	Write(String)
+}
+
+struct Shallow;
```
jonhoo: Might as well `#[derive(ShallowCopy)]` for this as well, no?

evmap-derive/tests/lib.rs
```diff
@@ -0,0 +1,28 @@
...
+
+impl evmap::ShallowCopy for Shallow {
+	unsafe fn shallow_copy(&self) ->  std::mem::ManuallyDrop<Self> {
+		unimplemented!();
+	}
+}
+
+#[derive(ShallowCopy)]
```
jonhoo: We should probably also have some tests for tuple structs and for generic types. It'd also be good to have some `compile_fail` tests that check that you _cannot_ derive `ShallowCopy` for types that contain non-`ShallowCopy` types.

tests/lib.rs
```diff
@@ -80,49 +80,48 @@ fn mapref() {
     // get a read ref to the map
     // scope to ensure it gets dropped and doesn't stall refresh
     {
-        let map = r.read();
+        let map = r.read().unwrap();
```
jonhoo:
```suggestion
        let map = r.read().expect("the map has not been destroyed");
```

tests/lib.rs
```diff
@@ -147,21 +146,17 @@ fn mapref() {
...
 
     drop(w);
     {
         let map = r.read();
-        assert!(map.is_destroyed());
-        assert!(map.is_empty());
-        assert!(!map.contains_key(&x.0));
-        assert!(map.get(&x.0).is_none());
-        assert!(map.meta().is_none());
+        assert!(map.is_none());
```
jonhoo:
```suggestion
        assert_eq!(map, None, "the map should have been destroyed");
```

jonhoo:
I think `get_owned` is probably overly specialized, and can be left out :)

MayaWolf:
I think that should cover all of your feedback so far!

evmap-derive/tests/failing/lib.rs
```diff
@@ -0,0 +1,25 @@
...
+struct Broken {
+	f1: i32,
+	f2: std::cell::Cell<()>,
+}
+
+#[derive(ShallowCopy)]
+struct AlsoBroken(i32, std::cell::Cell<()>);
+
+#[derive(ShallowCopy)]
+struct Generic<T> {
```
jonhoo: Hmm, why should this not work? Seems reasonable to me as long as the `T: ShallowCopy` bound is added?

jonhoo: In fact, don't we have the _exact_ same test passing in the other test file?

jonhoo:
Yup, I think we're very close now! The negative tests are, I think, the only thing really missing. Plus metadata and docs for the `evmap-derive` sub-crate. Metadata I can take care of after merging though. We should probably also update the docs for `evmap::ShallowCopy` to mention `evmap_derive`.

As an aside, what do you think of naming the derive crate `shallowcopy_derive`?

evmap-derive/tests/failing/lib.rs
MayaWolf: We do, and it does! It's just there as a setup for the failing usage in the function below. The derive automatically adds that bound, which is pretty much just what I'm testing here.

MayaWolf:
What kind of negative tests do you mean? Isn't that what's in `failing` now?

In terms of docs, are you just looking for a README.md, or something more in-depth? I'm not really sure what I'd document since it's "just a derive macro"; it doesn't do anything unusual and offers no configuration - but anything specific you'd like, I can add.

I personally think `evmap-derive` (dash or underscore is up to you; they seem to be about equally common) is a more useful name since it's entirely tied to evmap (unless you'd like to extract `ShallowCopy` into its own crate as well, but I think that would be overkill), and even though there's only one for now, it's ultimately a crate for "evmap's derive macros".

evmap-derive/tests/failing/lib.rs
jonhoo: Ah, you're right, I missed that the error is on _use_, not on declaration. Ignore me — clearly today is not my day :sweat_smile:

jonhoo:
Sorry, by "missing" I meant "remaining", by which I meant "my outstanding comments". But since that was just me being stupid, I think we're good on that front!

I was thinking `#![deny(missing_docs)]`, which will require you to write `//!` docs in `evmap-derive/src/lib.rs`. In those, I think we want to describe roughly what it does ("why would you want this"), some quick usage examples (really just an enum, a struct, and a generic type that all have `derive(ShallowCopy)`), and a link to the docs for `ShallowCopy`. I'd also mention for generic types that the derive adds a `ShallowCopy` bound for all of them, and that if your type does not directly contain a given generic parameter (e.g., it contains `Box<T>`), then you will have to implement `ShallowCopy` yourself. A "this is what the derive produces" for something like
```rust
#[derive(ShallowCopy)]
struct Generic<T> { field: T }
```
Often goes a _long_ way for users to get a good grasp on what's going on and how to use it correctly.

I have actually considered extracting `ShallowCopy` into its own crate, which is why I'm asking, but I think I agree with you that `evmap-derive` is good for now. I doubt we'll have other derives any time soon, but I guess it doesn't hurt to name for that case.

MayaWolf:
Added the derive docs! Lemme know if you want any more changes, otherwise I'll go rebase it.

MayaWolf:
There's one more thing I wanted to ask you about, actually - for my purposes, it would be extremely helpful for the `ValuesInner::Long` backing store to be a `HashSet` rather than a `HashBag` - do you think that's something that could and should be integrated into the main repo behind a feature flag or other kind of branch (similar to how you've been doing with the `indexed` feature) or would you prefer I maintain my own fork for that?

evmap-derive/src/lib.rs
```diff
@@ -1,10 +1,57 @@
...
 use proc_macro2::{Ident, TokenStream};
 use quote::{format_ident, quote, quote_spanned};
 use syn::spanned::Spanned;
 use syn::{
     parse_macro_input, parse_quote, Data, DeriveInput, Fields, GenericParam, Generics, Index,
 };
 
+/// Implementation for `#[derive(ShallowCopy)]`
+///
+/// evmap provides the [ShallowCopy](evmap::shallow_copy::ShallowCopy) trait, allowing you to cheaply alias types that don't otherwise implement `Copy`.
```
jonhoo:
```suggestion
/// evmap provides the [`ShallowCopy`](evmap::shallow_copy::ShallowCopy) trait, which allows you to cheaply alias types that don't otherwise implement `Copy`.
```

evmap-derive/src/lib.rs
```diff
@@ -1,10 +1,57 @@
...
+/// Basic implementations are provided for common types such as `String` and `Vec`, but it must be implemented manually for structs using these types.
+///
+/// This macro attempts to simplify this task. It only works on types whose members all implement `ShallowCopy`.
+/// If this is not possible, consider using [CopyValue](evmap::shallow_copy::CopyValue), `Box`, `Arc` instead.
```
jonhoo:
```suggestion
/// If this is not possible, consider using [CopyValue](evmap::shallow_copy::CopyValue), `Box`, or `Arc` instead.
```

evmap-derive/src/lib.rs
```diff
@@ -1,10 +1,57 @@
...
+///
+/// #[derive(ShallowCopy)]
+/// struct Generic<T> { field: T }
+///
+/// #[derive(ShallowCopy)]
+/// enum Things<T> { One(Thing), Two(Generic<T>) }
+/// ```
+///
+/// ## Generated implementations
+/// For any type, the generated implementation calls [shallow_copy](evmap::shallow_copy::ShallowCopy::shallow_copy), removing the inner `ManuallyDrop` wrappers.
```
jonhoo:
```suggestion
/// The generated implementation calls [`shallow_copy`](evmap::shallow_copy::ShallowCopy::shallow_copy) on all the members of the type, and lifts the `ManuallyDrop` wrappers to the top-level return type.
```

evmap-derive/src/lib.rs
```diff
@@ -1,10 +1,57 @@
...
+///
+/// Given generic types, it will add `ShallowCopy` bounds to all type parameters.
```
jonhoo:
```suggestion
/// For generic types, the derive adds `ShallowCopy` bounds to all the type parameters.
```

evmap-derive/src/lib.rs
```diff
@@ -1,10 +1,57 @@
...
+///
+/// For instance, for the following code...
+/// ```
+/// # use evmap_derive::ShallowCopy;
+/// #[derive(ShallowCopy)]
+/// struct Generic<T> { field: T }
+/// ```
+/// ...this will generate...
```
jonhoo:
```suggestion
/// ...the derive generates...
```

jonhoo:
Docs look pretty good -- I just left some minor nits. It'd be good to also update the `ShallowCopy` docs to mention `evmap-derive` for users who have more complex values.

I'm not sure I follow how it could be a `HashSet` instead of a `HashBag` and still retain the same semantics for `Short` and `Long`? It would mean that once you go over a certain length, duplicates start disappearing...

MayaWolf:
As I'm understanding it, anything above length 1 uses hashbag instead of smallvec, correct?

Given that, the "remove duplicates" functionality is actually exactly what I'm looking for, since at least in my use case, I'm exclusively dealing with uniques, and at-least-once delivery. Of course, it's very much something that users would need to opt into using a feature flag or generics - if you're even interested in supporting this functionality at all.

jonhoo:
@MayaWolf Ah, no, anything above [`BAG_THRESHOLD`](https://github.com/jonhoo/rust-evmap/blob/65771693a361cf3dfba5e52072bd8c8605e09459/src/values.rs#L6) becomes a `HashBag`, so you'd then have a data structure where duplicates are only removed once there are at least that many things in the bag. Out of curiosity -- what does having duplicate elimination buy you, when the bag essentially gives you the same thing? Is it because you want to explicitly be able to iterate over distinct values or something?

jonhoo:
Also, if all the values are unique anyway, why does it matter whether it's a `HashSet` or a `HashBag`? They should then be pretty much equivalent.

src/shallow_copy.rs
```diff
@@ -26,6 +26,9 @@ use std::ops::{Deref, DerefMut};
 /// If you have a non-`Copy` type, the value returned by `shallow_copy` should point to the same
 /// data as the `&mut self`, and it should be safe to `mem::forget` either of the copies as long as
 /// the other is dropped normally afterwards.
+///
+/// For structs that exclusively consist of types implementing `ShallowCopy`, consider using the `evmap-derive` crate, which contains a derive macro for `ShallowCopy`.
```
jonhoo:
```suggestion
/// For complex, non-`Copy` types, you can place the type behind a wrapper that implements `ShallowCopy` such as `Box` or `Arc`. Alternatively, if your type is made up of types that all implement `ShallowCopy`, consider using the `evmap-derive` crate, which contains a derive macro for `ShallowCopy`. See that crate's documentation for details.
```

jonhoo: Also, line-wrapping.

MayaWolf:
Iteration is part of it, but it's especially relevant when _removing_ a value, and making sure it's then actually gone (and doesn't happen to have been in there twice due to at-least-once delivery). Basically, right now, I have to check in a lot of places before inserting whether the value already exists, to make sure everything else is stilll behaving as expected.

src/read/read_ref.rs
```diff
@@ -65,23 +62,34 @@ where
     /// Note that not all writes will be included with this read -- only those that have been
     /// refreshed by the writer. If no refresh has happened, or the map has been destroyed, this
     /// function returns `None`.
-    pub fn get<'a, Q: ?Sized>(&'a self, key: &'_ Q) -> Option<&'a Values<V, S>>
+    pub fn get<Q: ?Sized>(&'rh self, key: &'_ Q) -> Option<&'rh Values<V, S>>
```
jonhoo: Looking at this (and `get_one`) again, I don't _think_ this should be tied to `'rh`. I mean, it _can_ be, but what matters is that the return value is tied to the lifetime of `&self`. This updated form requires that the reference to the handle live as long as the `ReadGuard`'s borrow, which isn't a necessary requirement. This should therefore probably stay the way it was, as that is the most relaxed bound that is still safe. And `get_one` should have the same.

jonhoo:
@MayaWolf How about adding a `remove_all`, which returns a `usize` of how many values where removed? I think that would let you continue to use `HashBag` while also supporting your use-case? For iteration it's trickier since you'd have to also eliminate duplicates from the `SmallVec` to support that...

MayaWolf:
Hmm...if we could make `BAG_THRESHOLD` configurable, add a `remove_all`, and indirect access to HashBag's `set_iter`, that could work just as well!

I say make it configurable since I don't think it would make sense to use smallvec for anything above size 1 in that use case, since otherwise they'd just need to be temporarily hashed in the iterator anyway.

jonhoo:
Hmm, making `BAG_THRESHOLD` should be doable I think. There'd be a question around how exactly we do that, but a simple method on `WriteHandle` might be sufficient. Even changing it dynamically should be possible, with a note that existing values won't immediately change to match the new limit.

Forwarding access to `set_iter` is more complicated, since it only makes sense (I think) if `BAG_THRESHOLD == 1`. One way to go about it would be to add another type parameter to all the evmap types that is either `Bag` or `Set` (both are unit types, `Bag` is the default). `set_bag_threshold` would then only be available on `WriteHandle<..., Bag>`, and the constructor that produces `WriteHandle<..., Set>` always sets the threshold to 1. Then you'd have two `ReadHandle::iter` impls, one for `ReadHandle<..., Set>` and one for `ReadHandle<..., Bag>`. The former uses `set_iter` (and asserts all the counts are 1), the latter just calls `iter` as usual. You may also need to specialize insert for `Set` so that it eliminates duplicates, which may be a bit of a pain.

Not sure all of ^ would be worth it. Another option is to make the variants of `enum ValuesInner` public (essentially by removing the `Values` wrapper type), so that users can directly operate on `&SmallVec` and `&HashBag` on their own for reads? You'd still need to add `remove_all`, but at least it would take care of the iterator issue.

MayaWolf:
I'm curious what you think about this one! I went for the second option you proposed, exposing the inner value storage directly. It allows the user to specify which inner value storage they would like to use - with the built-in choices currently being `HashBag`, `HashSet` and `Option`.

The changeset for this is pretty large, but that's mostly because I had to change generic parameters in a lot of places.

I also removed some of the stuff I added in previous commits, as it wasn't really applicable or necessary anymore. In the case of `Option`, the value is flattened before being exposed to the user, so single values can be worked with conveniently.

I tested this using both `cargo test` and a production application, and it seems to be working well!

Let me know what you think about this and, if you like it, what you'd like me to add!

Oh, one more thing: The implementation of `Values` for `HashBag` currently doesn't pass the hasher along, as `Debug` is only implemented for the version with `RandomState` - that's probably something that would need to be corrected in that crate.

jonhoo:
Ah, hmm, no, that wasn't quite what I had in mind with my second suggestion. It was much simpler: just replace `Values` with `ValuesInner`, so that when the user gets a `Values` they are able to directly operate on the contained `SmallVec` or `HashBag`. No generics should be needed for that.

I think the proposed change overcomplicates the interface beyond what I'd like. It also makes some of the operations pretty weird in terms of semantics. `push` if `Values` is an `Option` is now really a `replace` for example. It would also mean that we lose the ability to dynamically switch between a `SmallVec` and a `HashBag`, unless we introduce yet another `impl Values` type that does things dynamically.

I think, for what you need, these are the modifications to make:
 - Add `remove_all`
 - Replace `Values` with `ValuesInner` (basically, remove the wrapper)
 - Make `BAGIFY_THRESHOLD` configurable.

Your code should then have all it needs to work only with values with distinct sets (I think).

MayaWolf:
My thoughts were that `replace` could be removed as a separate function, since to my knowledge it's only really useful for single-value maps - which would be more directly covered by the `Option` impl. Generally, single-value maps wouldn't need any special handling in the API anymore (so I actually think it would ultimately simplify it).

I did omit the impl for dynamic smallvec+hashbag since I wasn't sure how necessary it would be (the user could just choose an optimal value-bag type themselves) but it could be added very easily to use as a reasonable default.

Overall, I do think this has potential for making things easier and more versatile to work with, but ultimately it's of course entirely up to you; if you'd prefer me to revert the last part I will.

jonhoo:
I think we maybe talked past each other — my worry with your proposed approach is that the semantics of each `Operation` variant now depends on what `V: Values` you have. `Push` does not _mean_ the same thing if `V = Option<T>` and `V = SmallVec<T>`. And, sadly, we cannot make the variants of `Operation` depend on the `V` (unless some kind of associated type trickery...). I suppose maybe with this approach we could have different methods on `impl WriteHandle<..., Option<T>>` and `impl WriteHandle<...., SmallVec<T>>` though, which might help?

I think what I'd prefer is to land this PR without the last commit, and then we open a new PR where we see if we can find a way to make that more sweeping change in isolation :)

MayaWolf:
Makes a lot of sense, yeah! This one should be ready for merge now; I'll open another PR once it is!

jonhoo:
Also, quick not on commit messages: it's usually best to keep the first line brief (50 chars), so that it does not get cut off in various UIs (like GitHub's). Then an empty line, and then a more detailed explanation of the change wrapped at 72 chars. There are a lot of resources on line on exact how/why (like [this one](https://chris.beams.io/posts/git-commit/)), but at least keeping the first line short is nice for GitHub specifically!

MayaWolf:
Hi! Sorry this took me so long to get back to, I haven't really been able to work on Rust stuff at all for a little while.

I've been trying a few ways of getting CI to work; unfortunately I've never worked with Azure pipelines before so I'm not super familiar with how the setup works there. When I try to simply add
```
 - template: default.yml@templates
   parameters:
     minrust: 1.40.0
     dir: evmap-derive
```
below the existing lines, it complains about duplicated job names, and with the way I've set it up now, it complains about the format of `default.yml` - I'm really not familiar enough with Azure to fix this, and unfortunately don't have the time right now to do in-depth research about it...do you know what the quick fix would be here?

jonhoo:
No worries at all — life takes precedence :)

Ah, yes, that's a little awkward. I think I'd just bypass the template and add:
```yaml
 - job: derive
   displayName: "Test evmap-derive"
   pool:
     vmImage: ubuntu-latest
   steps:
     - template: install-rust.yml@templates
     - script: cargo test
       displayName: cargo test
```
Probably just above the `benchmark` job here:
https://github.com/jonhoo/rust-evmap/blob/6e935f2f23a2a13077778e37dc71a7b0d30a7210/azure-pipelines.yml#L6

src/read/read_ref.rs
```diff
@@ -70,18 +65,7 @@ where
...
-        }
-        inner.data.get(key).map(Values::user_friendly)
-    }
-
-    /// Returns true if the writer has destroyed this map.
-    ///
-    /// See [`WriteHandle::destroy`].
-    pub fn is_destroyed(&self) -> bool {
-        self.guard.is_none()
+        self.guard.data.get(key).map(Values::user_friendly)
```
jonhoo, line 70: Maybe I'm missing something, but I don't think the check for whether the map is ready now gets performed anywhere? It's not to my knowledge applied inside `handle` at least?

src/shallow_copy.rs
```diff
@@ -26,6 +26,10 @@ use std::ops::{Deref, DerefMut};
 /// If you have a non-`Copy` type, the value returned by `shallow_copy` should point to the same
 /// data as the `&mut self`, and it should be safe to `mem::forget` either of the copies as long as
 /// the other is dropped normally afterwards.
+///
+/// For complex, non-`Copy` types, you can place the type behind a wrapper that implements `ShallowCopy` such as `Box` or `Arc`.
```
jonhoo, line 30: This should probably be text wrapped like the rest of the docs above it.

jonhoo:
This is looking great — thanks for tidying up the commits, it made reviewing much easier. I left two relatively minor comments :)

jonhoo:
It looks like the `busybusybusy_fast` test is now racy, and occasionally [fails](https://dev.azure.com/jonhoo/jonhoo/_build/results?buildId=1252&view=logs&j=cc2e323d-a806-545f-bf6b-e1565f70f846&t=1b044aa0-9aef-56b9-bdd7-1f415a470aaf&l=67):
```
thread 'busybusybusy_fast' panicked at 'called `Result::unwrap()` on an `Err` value: Any', tests/lib.rs:284:9
```
The same [appears to be the case](https://dev.azure.com/jonhoo/jonhoo/_build/results?buildId=1252&view=logs&j=bc994297-029d-5157-6f73-4fe90c912368&t=f2b50a2b-32dd-5a67-fba7-e24e52af645e&l=113) for `busybusybusy_heap` (line 328) and `busybusybusy_slow` (also line 284).

I wasn't planning on raising this, but since some last changes may have to be made, I think the docs in `evmap-derive` also could use text wrapping. And while we're on the topic, I noticed you sometimes you single line breaks in the docs; these (I believe) do not end up making it into the final documentation. You have to use a double-line break to make a new paragraph. See for example the text you recently wrapped in `shallow_copy`. If you don't want a paragraph break, it's usually nicer to just not have the single line break there at all.

src/read/mod.rs
```diff
@@ -298,7 +326,23 @@ where
...
+    pub fn contains_value<Q: ?Sized, W: ?Sized>(&self, key: &Q, value: &W) -> bool
+    where
+        K: Borrow<Q>,
+        V: Borrow<W>,
+        Q: Hash + Eq,
+        W: Hash + Eq,
+    {
+        self.get_raw(key.borrow())
+            .map(|x| x.user_friendly().contains(value))
+            .unwrap_or_default()
```
jonhoo, line 345: Is `unwrap_or(false)` clearer here perhaps?

MayaWolf:
Let me know whether you're happy with this!

jonhoo:
This is great, thank you for sticking with it! :tada:

Issue #31, opened 2019-06-07
Bring up test coverage


jonhoo:
Write some test cases (even simple ones are welcome!) to help increase [code coverage](https://codecov.io/gh/jonhoo/rust-evmap/tree/master/src).

jonhoo:
A lovely way to improve here would be to combine the tarpaulin coverage results for `left-right` and `evmap`, and to expand on `left-right`'s test suite as well. We should also probably adjust CI to run `left-right` itself through miri/sanitizers/etc.

PR #56, opened 2020-06-27
Adds tests for get_one

mlodato517:
**This PR**
Attempting to help along with https://github.com/jonhoo/rust-evmap/issues/31.

**Notes**
It appeared that [`get_one`](https://codecov.io/gh/jonhoo/rust-evmap/src/master/src/read/mod.rs#L291) wasn't tested. I'm not sure if these tests are in the right place or if they should live in `read/mod`. Or maybe this isn't tested because it's not valuable functionality? 🤷

I tried to handle both inner cases - `ValuesInner::Short` and `ValuesInner::Long`. I'm not 100% sure it's necessary to test `ValuesInner::Long` because, although it's supported, it's clearly documented that, in the case of multiple values, the exact return value is not guaranteed. So it's nice to make sure it doesn't explode with multiple values but it also doesn't feel particularly useful asserting any particular returned value. I also feel pretty badly about the magic number with no hard-reference to `BAG_THRESHOLD`. ~If we do want to improve the readability there, it looked like there was a `with_capacity` on `Options` but I didn't see one for `evmap` so I wasn't sure how to do that idiomatically. I probably read the docs wrong.~ It looks like you can do something like:
```rust
Options::default().with_capacity(BAG_THRESHOLD).construct()
```
although maybe there's a slightly more compact method.

~Finally, I don't know how to check code coverage before merging this so acknowledging that there's no guarantee the coverage will increase, I'm fine with just chucking this PR (but would still like to help if any more specific guidance is available!).~ Looks like it shows up right there in the PR! Neat!

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/rust-evmap/56)

### DIFF
--- a/src/values.rs
+++ b/src/values.rs
@@ -1,8 +1,10 @@
-use crate::BAG_THRESHOLD;
 use std::borrow::Borrow;
 use std::fmt;
 use std::hash::{BuildHasher, Hash};
 use std::mem::ManuallyDrop;
+
+/// This value determines when a value-set is promoted from a list to a HashBag.
+const BAG_THRESHOLD: usize = 32;
 
 /// A bag of values for a given key in the evmap.
 #[repr(transparent)]
--- a/src/values.rs
+++ b/src/values.rs
@@ -1,9 +1,8 @@
+use crate::BAG_THRESHOLD;
 use std::borrow::Borrow;
 use std::fmt;
 use std::hash::{BuildHasher, Hash};
 use std::mem::ManuallyDrop;
-
-const BAG_THRESHOLD: usize = 32;
 
 /// A bag of values for a given key in the evmap.
 #[repr(transparent)]
--- a/src/values.rs
+++ b/src/values.rs
@@ -308,14 +308,106 @@ where
 #[cfg(test)]
 mod tests {
     use super::*;
+    use std::collections::hash_map::RandomState;
+
+    macro_rules! assert_empty {
+        ($x:expr) => {
+            assert!($x.is_empty());
+            assert_eq!($x.len(), 0);
+            assert_eq!($x.iter().count(), 0);
+            assert_eq!($x.into_iter().count(), 0);
+            assert_eq!($x.get_one(), None);
+        };
+    }
+
+    macro_rules! assert_len {
+        ($x:expr, $n:expr) => {
+            assert!(!$x.is_empty());
+            assert_eq!($x.len(), $n);
+            assert_eq!($x.iter().count(), $n);
+            assert_eq!($x.into_iter().count(), $n);
+        };
+    }
+
+    fn has_short_values<T, S>(v: &Values<T, S>) -> bool {
+        match v.0 {
+            ValuesInner::Short(_) => true,
+            _ => false,
+        }
+    }
+    fn has_long_values<T, S>(v: &Values<T, S>) -> bool {
+        !has_short_values(v)
+    }
 
     #[test]
     fn sensible_default() {
         let v: Values<i32> = Values::default();
-        assert!(v.is_empty());
-        assert_eq!(v.len(), 0);
+        assert!(has_short_values(&v));
         assert_eq!(v.capacity(), 1);
-        assert_eq!(v.iter().count(), 0);
-        assert_eq!(v.into_iter().count(), 0);
+        assert_empty!(v);
+    }
+
+    #[test]
+    fn short_values() {
+        let hasher = RandomState::default();
+        let mut v = Values::new();
+
+        let values = 0..BAG_THRESHOLD - 1;
+        let len = values.clone().count();
+        for i in values.clone() {
+            v.push(i, &hasher);
+        }
+
+        for i in values.clone() {
+            assert!(v.contains(&i));
+        }
+        assert!(has_short_values(&v));
+        assert_len!(v, len);
+        assert_eq!(v.get_one(), Some(&0));
+
+        v.clear();
+
+        assert_empty!(v);
+
+        // clear() should not affect capacity or value type!
+        assert!(v.capacity() > 1);
+        assert!(has_short_values(&v));
+
+        v.shrink_to_fit();
+
+        assert_eq!(v.capacity(), 1);
+    }
+
+    #[test]
+    fn long_values() {
+        let hasher = RandomState::default();
+        let mut v = Values::new();
+
+        let values = 0..BAG_THRESHOLD;
+        let len = values.clone().count();
+        for i in values.clone() {
+            v.push(i, &hasher);
+        }
+
+        for i in values.clone() {
+            assert!(v.contains(&i));
+        }
+        assert!(has_long_values(&v));
+        assert_len!(v, len);
+        assert!(values.contains(v.get_one().unwrap()));
+
+        v.clear();
+
+        assert_empty!(v);
+
+        // clear() should not affect capacity or value type!
+        assert!(v.capacity() > 1);
+        assert!(has_long_values(&v));
+
+        v.shrink_to_fit();
+
+        // Now we have short values!
+        assert!(has_short_values(&v));
+        assert_eq!(v.capacity(), 1);
     }
 }
--- a/src/values.rs
+++ b/src/values.rs
@@ -319,8 +319,8 @@ mod tests {
 
     macro_rules! assert_empty {
         ($x:expr) => {
-            assert!($x.is_empty());
             assert_eq!($x.len(), 0);
+            assert!($x.is_empty());
             assert_eq!($x.iter().count(), 0);
             assert_eq!($x.into_iter().count(), 0);
             assert_eq!($x.get_one(), None);
@@ -329,8 +329,8 @@ mod tests {
 
     macro_rules! assert_len {
         ($x:expr, $n:expr) => {
-            assert!(!$x.is_empty());
             assert_eq!($x.len(), $n);
+            assert!(!$x.is_empty());
             assert_eq!($x.iter().count(), $n);
             assert_eq!($x.into_iter().count(), $n);
         };
--- a/src/values.rs
+++ b/src/values.rs
@@ -100,6 +100,11 @@ impl<T, S> Values<T, S> {
             ValuesInner::Long(ref v) => v.contains(value) != 0,
         }
     }
+
+    #[cfg(test)]
+    fn is_short(&self) -> bool {
+        matches!(self.0, ValuesInner::Short(_))
+    }
 }
 
 impl<'a, T, S> IntoIterator for &'a Values<T, S> {
@@ -331,20 +336,10 @@ mod tests {
         };
     }
 
-    fn has_short_values<T, S>(v: &Values<T, S>) -> bool {
-        match v.0 {
-            ValuesInner::Short(_) => true,
-            _ => false,
-        }
-    }
-    fn has_long_values<T, S>(v: &Values<T, S>) -> bool {
-        !has_short_values(v)
-    }
-
     #[test]
     fn sensible_default() {
         let v: Values<i32> = Values::default();
-        assert!(has_short_values(&v));
+        assert!(v.is_short());
         assert_eq!(v.capacity(), 1);
         assert_empty!(v);
     }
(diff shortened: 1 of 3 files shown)


tests/lib.rs
```diff
@@ -732,3 +732,40 @@ fn retain() {
...
+
+    let (r, mut w) = evmap::new();
+
+    w.insert(x.0, x);
+
+    assert_match!(r.get_one(&x.0), None);
+
+    w.refresh();
+
+    assert_match!(r.get_one(&x.0).unwrap().as_ref(), ('x', 42));
```
mlodato517, line 748: Is this the right way to read the inner value?

tests/lib.rs
```diff
@@ -732,3 +732,40 @@ fn retain() {
...
+}
+
+#[test]
+fn get_one_long() {
+    let x = ('x', 42);
```
mlodato517, line 753: I kept this value this way for consistency but didn't want to just add this same value every time. Wanted to sort of document the fact that the output value is not determined. Am open to:
1. inserting `x` every time
1. changing/removing `x` because we don't need the tuple
1. anything else!

tests/lib.rs
```diff
@@ -732,3 +732,40 @@ fn retain() {
...
+
+    let (r, mut w) = evmap::new();
+
+    // Add 32 items to meet the BAG_THRESHOLD and
```
mlodato517, line 757: Gross. I hate relying on this implementation detail here. Wouldn't want this test to lose its flair if `BAG_THRESHOLD` changes. And the worst part is it'll just silently keep passing. So I'm willing to bet this test just isn't worth having in here?

tests/lib.rs
jonhoo: I think you could probably use
```rust
assert_match!(r.get_one(&x.0).as_deref(), Some(('x', 42)));
```
instead, but either is fine!

tests/lib.rs
jonhoo: I would probably change this to `let key = 'x';` :)

tests/lib.rs
jonhoo: I think what I would do here is make `BAG_THRESHOLD` be `pub`, but `#[doc(hidden)]`, and then add to its doc comment that it is only `pub` to be used in tests.

jonhoo:
This is great, thank you! I left some comments :)

tests/lib.rs
mlodato517: Great! I was afraid of exposing that but `doc(hidden)` seems like a good way to expose it without encouraging its use 👍 Two quick questions on this:

1) I've never used `rustdoc` before. To add the attribute _and_ the rustdoc is it like:
```rs
/// Comment here
#[doc(hidden)]
pub ...
```
? Or are the comments and attribute switched? Or is it some other configuration? In all the examples I see it's like:
```rs
// Non-doc-comment here
#[doc(hidden)]
pub ...
```

2) I can make `BAG_THRESHOLD` `pub` but to access it in `tests/lib.rs` I think I have to also make the [`values` module public](https://github.com/jonhoo/rust-evmap/blob/master/src/lib.rs#L248) or add it to the [`pub use` for `values`](https://github.com/jonhoo/rust-evmap/blob/master/src/lib.rs#L249). I'm assuming of those two, the second is preferable but I'm just checking first in case neither are okay!

jonhoo: Ah, for 1, I think the order actually doesn't matter. Though my personal preference is doc comment, then attributes.

For 2, I'd just move the definition of `BAG_THRESHOLD` to the root of the crate!

tests/lib.rs
mlodato517: Great idea, I like that much better! Changes in ac34710870690a8d5ea493ccd9feb6749d324626. I didn't use this syntax in the other test so I could still use `contains` on the range. I could probably use `map(Some).any?` or something but I didn't know if that was worth it.

tests/lib.rs
mlodato517: Much more readable! Changes in f804389bf60096f8d563173a6c143ad4ce30557b

tests/lib.rs
mlodato517: Awesome, thanks for all that help. Changes in 40062d0c28bb844163fdb6e63e3dbfa602a023d0. I was a little uncertain if I should write better comments for `BAG_THRESHOLD` since it's defined farther from where it's used but felt silly writing more doc comments that were hidden. I kind of wanted to write non-doc comments 🤔 Which kind of matches the other examples I see so now I understand that. But I went with what we'd already mentioned and am quite willing to update it any which way!

tests/lib.rs
```diff
@@ -732,3 +732,39 @@ fn retain() {
...
+    assert_match!(r.get_one(&x.0).as_deref(), Some(('x', 42)));
+}
+
+#[test]
+fn get_one_long() {
+    let x = 'x';
+
+    let (r, mut w) = evmap::new();
+
+    // Add BAG_THRESHOLD items to ensure the inner type is ValuesInner::Long.
```
jonhoo, line 757: It might be good to test if the value actually got promoted to a `ValuesInner::Long` here. That may require another `#[doc(hidden)]` method.

mlodato517: Haha I was also struggling with that idea and that would be a reasonable reason why :-)

tests/lib.rs
jonhoo: I think adding a bit more of a generous doc comment would be good. Something like:

```rust
/// This value determines when a value-set is promoted from a list to a HashBag.
```

tests/lib.rs
mlodato517: Hmm actually, since we want to test this anyway, would it be better if I structured the tests like:
```rust
// tests/lib.rs
fn test_get_one() {
  // ensure get_one gets one value from the underlying values
  // whatever that means
}
```
```rust
// src/values.rs
fn test_value_type_upgrade() {
  // ensure the value type stays short until we exceed
  // the threshold
}

fn test_get_one() {
  // ensure get_one works regardless of value type
}
```
?

jonhoo: Yes, I agree with you, these should probably be unit tests instead.

Having two tests seems fine, though you'll still want to rely on the particular threshold in the larger test.

mlodato517: Okay, I'll shoot for unit tests 👍

> though you'll still want to rely on the particular threshold in the larger test.

Quick question - which is "the larger test"? Sorry for all the back and forth this tiny PR has caused 😞 I'm terrified of the dollar-per-value here haha

jonhoo: Oh, sorry, that was vague of me. I expect that `test_value_type_upgrade` will _just_ test that the change happens as expected at the threshold, whereas `test_get_one` will be "larger", and will also test _behavior_ after that threshold is reached.

And don't worry about it at all, I'm happy to help :)

src/values.rs
```diff
@@ -309,14 +310,106 @@ where
...
+    macro_rules! assert_len {
+        ($x:expr, $n:expr) => {
+            assert!(!$x.is_empty());
+            assert_eq!($x.len(), $n);
+            assert_eq!($x.iter().count(), $n);
+            assert_eq!($x.into_iter().count(), $n);
+        };
+    }
+
+    fn has_short_values<T, S>(v: &Values<T, S>) -> bool {
```
mlodato517, line 334: I don't have to do this this way. I could add them in the `impl` outside the test module. I could also add them to an `impl` inside the test module so they're only available for tests.

I could also, instead of checking `v.0` in this method, I could use the public `v.iter()` in a test to do like:
```rust
match v.iter() {
  ValuesIter::Short(_) => {},
  _ => panic!("Does not have short values!"),
}
```
because `panic!` and `assert!` are basically the same. Lemme know which you'd prefer 👍 (or whatever other alternative!)

src/values.rs
```diff
@@ -309,14 +310,106 @@ where
...
+        ($x:expr) => {
+            assert!($x.is_empty());
+            assert_eq!($x.len(), 0);
+            assert_eq!($x.iter().count(), 0);
+            assert_eq!($x.into_iter().count(), 0);
+            assert_eq!($x.get_one(), None);
+        };
+    }
+
+    macro_rules! assert_len {
```
mlodato517, line 325: I could make `assert_empty` a special case of `assert_len` but I didn't think that was quite as readable. Open to other options though!

tests/lib.rs
mlodato517: I ended up making this private again since I moved the tests to `src/values`. I hope that's okay. I left in the doc comment 🤷. e9bd66964fcca1221bbb6b3c42e22e8f8246b062

tests/lib.rs
mlodato517: 30f33489d114b6354da9d37bcaa2340501498b7e

src/values.rs
```diff
@@ -308,14 +310,106 @@ where
 #[cfg(test)]
 mod tests {
     use super::*;
+    use std::collections::hash_map::RandomState;
+
+    macro_rules! assert_empty {
+        ($x:expr) => {
+            assert!($x.is_empty());
+            assert_eq!($x.len(), 0);
```
jonhoo, line 318: I would do the `assert_eq` on the length first, _then_ the check on `is_empty`. This is a bit more helpful when debugging, because if the assertion fails you're told what the length _actually_ was. Same for `assert_len` below.

src/values.rs
jonhoo: No, that seems reasonable to me!

src/values.rs
jonhoo: I think a method on `Values` annotated with `#[cfg(test)]` is probably the way to go here

jonhoo: Also, you can simplify this implementation to:
```rust
fn is_short(&self) -> bool {
  matches!(self.0, ValuesInner::Short(_))
}
```

mlodato517: Ooo TIL! I was looking for something like this!

tests/lib.rs
jonhoo: Looks great!

src/values.rs
mlodato517: ac8fc8ca039225c21057adc41f83e6015879f600

src/values.rs
mlodato517: Good thinking! c907045225d49562e0e2c58bf8d4b4fc68350c02

mlodato517:
Also, slightly unrelated - I see the CI failing scarily (to someone who didn't know why). It seems like the version of `evmap` in `benchmark/Cargo.toml` can be updated to the current `11.0.0-alpha.1`. Is that something to fix or is that purposefully not being upgraded until a non-alpha/beta/whatever release lands?

jonhoo:
Oh, yeah, good catch! If you want to include a quick fix for that in this PR too, that'd be great.

mlodato517:
> Oh, yeah, good catch! If you want to include a quick fix for that in this PR too, that'd be great.

Sorry, just saw this. Do you mind if I make it a new PR? I already had a new PR ready to go with the changes and figured it'd keep each PR's focus clean? But I'm also 100% down to combine them to reduce overhead!

I could squash all the existing commits and then add the new commit that bumps the dependency version and not squash those two if you'd prefer 👍

jonhoo:
Beautiful! I merged that PR, and squashed and merged this one. Thanks for all your work on this :D

mlodato517:
It's been my pleasure! Are there any other tests I should focus on next? Or any other things you've been wanting to get to but haven't had time? I really want to start contributing and learning/writing rust. Seeing as I've already learned a handful of things from this, I'd say continuing on would be fruitful. So I'm ready and rearing for anything you need or know about someone else needing! 👍

jonhoo:
I think an awesome thing would be to adopt parts of the test suites from [`hashbrown`](https://github.com/rust-lang/hashbrown) and [`indexmap`](https://github.com/bluss/indexmap/). They have extensive test suites, which evmap could possibly benefit a lot from. You'll have to work around evmap having a slightly different API, but it shouldn't be too bad, at least for the simpler tests. Some [quicktests](https://github.com/jonhoo/griddle/blob/master/tests/quick.rs) would also be fantastic.

PR #58, opened 2020-07-01
Add some quicktests

mlodato517:
I realized I had some questions about which tests should do which things. I didn't know where to put the discussion so I opened a new PR. This PR is most definitely not ready for merging haha

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/rust-evmap/58)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,4 +31,6 @@ rand = { version = "0.7", default-features = false, features = ["alloc"], option
 slab = "0.4"
 
 [dev-dependencies]
+quickcheck = "0.9"
+quickcheck_macros = "0.9"
 rand = "0.7"
--- a/tests/lib.rs
+++ b/tests/lib.rs
@@ -748,3 +748,38 @@ fn get_one() {
 
     assert_match!(r.get_one(&x.0).as_deref(), Some(('x', 42)));
 }
+
+#[test]
+fn insert_remove() {
+    let x = 'x';
+
+    let (r, mut w) = evmap::new();
+
+    w.insert(x, x);
+
+    w.remove(x, x);
+    w.refresh();
+
+    // There are no more values associated with this key
+    assert!(r.get(&x).is_some());
+    assert_match!(r.get(&x).as_deref().unwrap().len(), 0);
+
+    // But the map is NOT empty! It still has an empty bag for the key!
+    assert!(!r.is_empty());
+    assert_eq!(r.len(), 1);
+}
+
+#[test]
+fn insert_empty() {
+    let x = 'x';
+
+    let (r, mut w) = evmap::new();
+
+    w.insert(x, x);
+
+    w.empty(x);
+    w.refresh();
+
+    assert!(r.is_empty());
+    assert!(r.get(&x).is_none());
+}
--- a/tests/quick.rs
+++ b/tests/quick.rs
@@ -1,0 +1,200 @@
+extern crate evmap;
+
+use evmap::{ReadHandle, WriteHandle};
+
+extern crate quickcheck;
+#[macro_use(quickcheck)]
+extern crate quickcheck_macros;
+
+use quickcheck::Arbitrary;
+use quickcheck::Gen;
+
+use rand::Rng;
+use std::cmp::{min, Ord};
+use std::collections::{HashMap, HashSet};
+use std::fmt::Debug;
+use std::hash::{BuildHasher, Hash};
+use std::ops::Deref;
+
+fn set<'a, T: 'a, I>(iter: I) -> HashSet<T>
+where
+    I: IntoIterator<Item = &'a T>,
+    T: Copy + Hash + Eq,
+{
+    iter.into_iter().cloned().collect()
+}
+
+#[quickcheck]
+fn contains(insert: Vec<u32>) -> bool {
+    let (r, mut w) = evmap::new();
+    for &key in &insert {
+        w.insert(key, ());
+    }
+    w.refresh();
+
+    insert.iter().all(|&key| r.get(&key).is_some())
+}
+
+#[quickcheck]
+fn contains_not(insert: Vec<u8>, not: Vec<u8>) -> bool {
+    let (r, mut w) = evmap::new();
+    for &key in &insert {
+        w.insert(key, ());
+    }
+    w.refresh();
+
+    let nots = &set(&not) - &set(&insert);
+    nots.iter().all(|&key| r.get(&key).is_none())
+}
+
+#[quickcheck]
+fn insert_empty(insert: Vec<u8>, remove: Vec<u8>) -> bool {
+    let (r, mut w) = evmap::new();
+    for &key in &insert {
+        w.insert(key, ());
+    }
+    w.refresh();
+    for &key in &remove {
+        w.empty(key);
+    }
+    w.refresh();
+    let elements = &set(&insert) - &set(&remove);
+    r.len() == elements.len()
+        && r.map_into::<_, Vec<()>, _>(|_, _| ()).len() == elements.len()
+        && elements.iter().all(|k| r.get(k).is_some())
+}
+
+use Op::*;
+#[derive(Copy, Clone, Debug)]
+enum Op<K, V> {
+    Add(K, V),
+    Remove(K),
+}
+
+impl<K, V> Arbitrary for Op<K, V>
+where
+    K: Arbitrary,
+    V: Arbitrary,
+{
+    fn arbitrary<G: Gen>(g: &mut G) -> Self {
+        match g.gen::<u32>() % 4 {
+            0 => Add(K::arbitrary(g), V::arbitrary(g)),
+            _ => Remove(K::arbitrary(g)),
+        }
+    }
+}
+
+fn do_ops<K, V, S>(ops: &[Op<K, V>], a: &mut WriteHandle<K, V, (), S>, b: &mut HashMap<K, Vec<V>>)
+where
+    K: Hash + Eq + Clone,
+    V: Clone + evmap::ShallowCopy + Eq + Hash,
+    S: BuildHasher + Clone,
+{
+    for op in ops {
+        match *op {
+            Add(ref k, ref v) => {
+                a.insert(k.clone(), v.clone());
+                b.entry(k.clone()).or_insert_with(Vec::new).push(v.clone());
+            }
+            Remove(ref k) => {
+                a.empty(k.clone());
+                b.remove(k);
+            }
+        }
+    }
+}
+
+fn assert_maps_equivalent<K, V, S>(a: &ReadHandle<K, V, (), S>, b: &HashMap<K, Vec<V>>) -> bool
+where
+    K: Clone + Hash + Eq + Debug,
+    V: Hash + Eq + Debug + Ord + Copy,
+    S: BuildHasher,
+{
+    assert_eq!(a.len(), b.len());
+    let keys = a.map_into::<_, Vec<K>, _>(|k, _| k.clone());
+    for key in &keys {
+        assert!(b.contains_key(key), "b does not contain {:?}", key);
+    }
+    for key in b.keys() {
+        assert!(a.get(key).is_some(), "a does not contain {:?}", key);
+    }
+    for key in &keys {
+        let mut ev_map_values: Vec<V> = a.get(key).unwrap().iter().copied().collect();
+        ev_map_values.sort();
+        let mut map_values = b[key].clone();
+        map_values.sort();
+        assert_eq!(ev_map_values, map_values);
+    }
+    true
+}
+
+/// quickcheck Arbitrary adaptor -- make a larger vec
+#[derive(Clone, Debug)]
+struct Large<T>(T);
+
+impl<T> Deref for Large<T> {
+    type Target = T;
+    fn deref(&self) -> &T {
+        &self.0
+    }
+}
+
+impl<T> Arbitrary for Large<Vec<T>>
+where
+    T: Arbitrary,
+{
+    fn arbitrary<G: Gen>(g: &mut G) -> Self {
+        let len = g.next_u32() % (g.size() * 10) as u32;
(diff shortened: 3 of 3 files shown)


Cargo.toml
```diff
@@ -31,4 +31,6 @@ rand = { version = "0.7", default-features = false, features = ["alloc"], option
 slab = "0.4"
 
 [dev-dependencies]
+quickcheck = "0.9"
+quickcheck_macros = "0.9"
```
mlodato517, line 35: It seemed from the docs that the macro is the "old way" and the attribute is the "new way". If that's not true or if we want to use the macro anyway, lemme know!

tests/lib.rs
```diff
@@ -748,3 +748,18 @@ fn get_one() {
...
+
+    let (r, mut w) = evmap::new();
+
+    w.insert(x, x);
+    w.refresh();
+
+    w.remove(x, x);
+    w.refresh();
+
+    assert!(r.is_empty());
```
mlodato517, line 764: I was copying the [quicktests from `indexmap`](https://github.com/bluss/indexmap/blob/master/tests/quick.rs#L67) and the `insert_remove` test failed. I made a small version here and this assertion fails. Unfortunately ... I can't tell if it's supposed to pass haha. If it is, I can work on getting this test to pass. If it's working correctly, then I can adjust the test to document that!

jonhoo: Hmm, that seems weird. Are you calling `remove`, `clear`, or `empty`? They all have slightly different semantics in evmap. Also, as always with evmap, keep in mind that nothing is visible to the reader until you've done a `refresh` on the `WriteHandle`!

Cargo.toml
jonhoo: I honestly don't know :sweat_smile:

tests/lib.rs
mlodato517: Good questions. Just `remove` for this test. And I do call `refresh` and see that the key no longer as associated values. But since the map still has a key (`remove` doesn’t remove keys) it isn’t considered empty and so neither is the overall `evmap` (I think).

It looks like `empty` would work because that calls `remove` on the inner map instead of the value of the key. But I guess I’m asking, “After calling `remove` on the last key-value pair, should the map be empty? Or should it retain the key for some purpose?” Once we know that, then we know if we need to update the test or update the code! :-)

jonhoo: So, `remove` removed _one_ value from a value-bag for a given key. Its semantics are (at least at the moment, and should probably remain) that it does not erase the key even when the last value is removed. That's what `empty` is for. Arguably the names could be better though :p

mlodato517: Haha I can throw up a quick PR to rename `remove` to `remove_value` and then a follow-up PR to rename `empty` to `remove` to make it match a regular map more closely. Or not!

tests/quick.rs
```diff
@@ -0,0 +1,200 @@
...
+        w.insert(key, ());
+    }
+    w.refresh();
+    for &key in &remove {
+        w.empty(key);
+    }
+    w.refresh();
+    let elements = &set(&insert) - &set(&remove);
+    r.len() == elements.len()
+        && r.map_into::<_, Vec<()>, _>(|_, _| ()).len() == elements.len()
```
mlodato517, line 63: I've been using `map_into` a lot. I realized that I could use `r.read().iter().flatten()`. I suppose that's clearer becase we can ditch a lot of the generic typing. But I wonder if it might be worth implementing `iter` or `keys` or something for the `ReadHandle`?

tests/quick.rs
```diff
@@ -0,0 +1,200 @@
...
+{
+    fn arbitrary<G: Gen>(g: &mut G) -> Self {
+        match g.gen::<u32>() % 4 {
+            0 => Add(K::arbitrary(g), V::arbitrary(g)),
+            _ => Remove(K::arbitrary(g)),
+        }
+    }
+}
+
+fn do_ops<K, V, S>(ops: &[Op<K, V>], a: &mut WriteHandle<K, V, (), S>, b: &mut HashMap<K, Vec<V>>)
```
mlodato517, line 87: Made it `Vec` out of laziness. Basically the evmap always has _some_ collection associated with the key. Figured it didn't matter which

tests/quick.rs
```diff
@@ -0,0 +1,200 @@
...
+                b.remove(k);
+            }
+        }
+    }
+}
+
+fn assert_maps_equivalent<K, V, S>(a: &ReadHandle<K, V, (), S>, b: &HashMap<K, Vec<V>>) -> bool
+where
+    K: Clone + Hash + Eq + Debug,
+    V: Hash + Eq + Debug + Ord + Copy,
```
mlodato517, line 110: This `Copy` and `Ord` are also laziness so I can easily sort and compare the values for each key between the maps. Am open to anything else!

tests/lib.rs
jonhoo: Hmm... That would be a backwards-incompatible change, which is a little sad. Maybe mark `empty` and `remove` as deprecated (with `#[deprecated]`), and add `remove_value` and `remove_entry`?

tests/quick.rs
```diff
@@ -0,0 +1,200 @@
...
+}
+
+#[quickcheck]
+fn operations_string(ops: Vec<Op<Alphabet, i8>>) -> bool {
+    let (r, mut w) = evmap::new();
+    let mut reference = HashMap::new();
+    do_ops(&ops, &mut w, &mut reference);
+    w.refresh();
+    assert_maps_equivalent(&r, &reference)
+}
```
mlodato517, line 200: Left out some tests like [keys and values iterators](https://github.com/jonhoo/griddle/blob/master/tests/quick.rs#L179) because evmap doesn't have them. If we want them, I'd be very willing to take a crack at those!

tests/lib.rs
mlodato517: Great idea! 👍

tests/quick.rs
jonhoo: Isn't `IntoIter` already implemented for `&ReadHandle`?

tests/quick.rs
jonhoo: No, that seems fine to me!

tests/quick.rs
mlodato517: Hmmmm maybe I'm just not seeing it? I'm only seeing iter methods on `MapReadRef` and `Values` (and `MapReadRef` does have an `into_iter` but it's the same as `iter` which is a little surprising 🤔).

jonhoo: Ohh, wait, sorry, I misread `ReadHandle`. I _don't_ think we want to impl `iter()` directly on `ReadHandle`, because it would make the call to `read()` implicit. That call holds up writers while it's active, so I'd prefer to have that call be as visible as possible.

mlodato517: Ahhh that makes oodles of sense. Which is why `map_into` is fine because it creates the iterator but consumes it immediately.

mlodato517: And we couldn't have a `keys` helper unless we require `K: Clone` or `K: Copy` because we can't give out a `Vec<&K>`/`Iterator<Item=&K>`, right? Or should we add an `impl` block for that case? It's probably only helpful for tests or other single-threaded applications in which case you probably wouldn't reach for this tool?

jonhoo: I think `keys` and `values` can (and probably should) exist on `MapReadRef`. I don't think we want to add any more methods to `ReadHandle` that take the guard behind the scenes. Even `map_into` is _arguably_ a foot-gun.

mlodato517: 👍 Roger that. I'd be glad to do that in a follow up PR! Don't want to scope creep this too much. And if we think these tests read all crappy then I can do a PR first for `keys` and `values` and then rebase this on top of that?

jonhoo: I think they're probably fine as-is, and then we can tidy them up with the follow-on PR?

mlodato517: Works for me!

tests/quick.rs
```diff
@@ -0,0 +1,200 @@
...
+    w.refresh();
+    let elements = &set(&insert) - &set(&remove);
+    r.len() == elements.len()
+        && r.map_into::<_, Vec<()>, _>(|_, _| ()).len() == elements.len()
+        && elements.iter().all(|k| r.get(k).is_some())
+}
+
+use Op::*;
+#[derive(Copy, Clone, Debug)]
+enum Op<K, V> {
```
jonhoo, line 69: If we wanted to be really fancy so that we could also throw `refresh` into this mix, we'd keep two reference maps to reflect "what readers see" separately from the "true" state of the map. And then `refresh` syncs the "true" state to the "reader" map by overwriting it with a clone.

mlodato517: Hmm you might've lost me there 🤔 Are you looking for:
1)
```rust
fn do_ops(ops: &[Op], a: &mut WriteHandle, b: &mut HashMap, c: &mut WriteHandle) {
    for op in ops {
        match *op {
            Add(ref k, ref v) => {
                a.insert(k.clone(), v.clone());
                b.entry(k.clone()).or_insert_with(Vec::new).push(v.clone());
                c.insert(k.clone(), v.clone());
            }
            Remove(ref k) => {
                a.empty(k.clone());
                b.remove(k);
                c.empty(k.clone());
            }
            Refresh() => {
              c.refresh()
           }
        }
    }
}
```
or
2)
```rust
fn do_ops(ops: &[Op], a: &mut WriteHandle, b: &mut HashMap, c: &mut WriteHandle) {
    for op in ops {
        match *op {
            Add(ref k, ref v) => {
                a.insert(k.clone(), v.clone());
                b.entry(k.clone()).or_insert_with(Vec::new).push(v.clone());
                c.insert(k.clone(), v.clone());
                c.refresh();
            }
            Remove(ref k) => {
                a.empty(k.clone());
                b.remove(k);
                c.empty(k.clone());
                c.refresh()
            }
        }
    }
}
```

And then is the test:
```rust 
#[quickcheck]
fn test(ops: Vec<Op<Alphabet, i8>>) -> bool {
    let (r, mut w) = evmap::new();
    let (other_r, mut other_w) = evmap::new();
    let mut reference = HashMap::new();
    do_ops(&ops, &mut w, &mut reference, &mut other_w);

    // assert r != other_r
    w.refresh()
    // assert r == other_r
```
?

jonhoo: None of those actually!
```rust
fn do_ops(ops: &[Op], w: &mut WriteHandle, r: &mut ReadHandle, wref: &mut HashMap, rref: &mut HashMap) {
    for op in ops {
        match *op {
            Add(ref k, ref v) => {
                wref.insert(k.clone(), v.clone());
                w.entry(k.clone()).or_insert_with(Vec::new).push(v.clone());
            }
            Remove(ref k) => {
                wref.remove(k);
                w.empty(k.clone());
            }
            Refresh => {
                w.refresh();
                *rref = wref.clone();
            }
        }
    }
}
```
And then in the assertion at the end, assert that `ReadHandle == rref`, and that after a `refresh`, `ReadHandle == wref`.

mlodato517: Very clever! Okay that makes lots of sense. I hope one day I make as much sense as you do!

mlodato517: And just to check, while I'm adding operations, I can't add a `RemoveValue` operation, can I? Because the arbitrary value probably won't be a value in the bag?

jonhoo: Oh, you totally can. Just make sure the value set isn't _too_ large. Something like `u8` or a limited alphabet tends to get you quite far.

mlodato517: Oh I didn't even notice we were already using `i8`. Considering the implementation of `Large` that's probably good enough?

jonhoo:
This is looking great, so I'll go ahead and merge it once CI is green! If you want to also submit those two follow-up PRs, that'd be awesome :D Thanks for your work on this!

tests/quick.rs
mlodato517: Added `RemoveValue` - 2da788840458579e4adad4358d455c1d1b6c2786

mlodato517:
> This is looking great, so I'll go ahead and merge it once CI is green! If you want to also submit those two follow-up PRs, that'd be awesome :D Thanks for your work on this!

No problem! I'll absolutely push those other PRs up sometime tomorrow! And I think we had a bit of a race condition - I pushed up the `RemoveValue` operation sometime around when you merged so I think that might've snuck in when you hadn't seen it. If there are any changes you want, lemme know and I'll just toss them in a new PR! Sorry about that timing

tests/quick.rs
```diff
@@ -109,6 +111,15 @@ fn do_ops<K, V, S>(
                 evmap.empty(k.clone());
                 write_ref.remove(k);
             }
+            RemoveValue(ref k, ref v) => {
+                evmap.remove(k.clone(), v.clone());
+                write_ref.get_mut(k).and_then(|values| {
+                    values
+                        .iter_mut()
+                        .position(|value| value == v)
+                        .and_then(|pos| Some(values.remove(pos)))
```
mlodato517, line 120: I guess I could `swap_remove` to speed up the tests since we sort later anyway

jonhoo: Oh, yeah, that'd be great! Unlikely to matter _too_ much though.

jonhoo:
Ooops, yes, I merged before you pushed that. Open another PR? :sweat_smile:

mlodato517:
100%! I’m a huge fan of the small PRs.

mlodato517:
@jonhoo Oh it [looks like my commit made it in before the merge](https://github.com/jonhoo/rust-evmap/blob/master/tests/quick.rs#L116). Do you want me to make any updates to it?

jonhoo:
Oh, nice, so much for race conditions. No, I think it's fine as-is! The change to `swap_remove` might be nice, but it isn't critical.

PR #59, opened 2020-07-07
Improve naming of remove methods

mlodato517:
**This PR**
- Deprecates the `remove` method for a more descriptive method `remove_value`
- Deprecates the `empty` method for a more description method `remove_empty`
- Uses `swap_remove` in a test where the order doesn't matter so why not

**Why?**
Using the standard lib as a North Star, [`remove`](https://doc.rust-lang.org/beta/std/collections/struct.HashMap.html#method.remove) should be a method that takes a key and removes that key from the map. In `evmap`, `remove` is meant to take a key _and a value_ and remove _just that value_ from the value bag for the key. To make that more clear, we deprecate `remove` and rename it to `remove_value`.

Unfortunately we can't then rename `empty` (`evmap`'s version of the std `HashMap`s `remove`) to `remove` for backwards compatibility. Instead, we rename it to `remove_entry`.

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/rust-evmap/59)

### DIFF
--- a/src/write.rs
+++ b/src/write.rs
@@ -748,4 +748,46 @@ mod tests {
         assert!(w.oplog[w.swap_index..].is_empty());
         w.flush();
     }
+
+    #[test]
+    fn flush_no_refresh() {
+        let x = 'x';
+
+        let (_, mut w) = new();
+
+        // Until we refresh, writes are written directly instead of going to the
+        // oplog (because there can't be any readers on the w_handle table).
+        w.refresh();
+
+        assert_eq!(w.second, true);
+
+        w.flush();
+
+        // No refresh because there are no operations
+        assert_eq!(w.second, true);
+
+        w.insert(x, x);
+        w.flush();
+
+        // A refresh happened!
+        assert_eq!(w.second, false);
+
+        // Reset this flag so we can see if another refresh happens
+        w.second = true;
+
+        w.flush();
+
+        // Subsequent flushes don't refresh until there are more ops!
+        assert_eq!(w.second, true);
+
+        w.insert(x, x);
+        w.flush();
+
+        assert_eq!(w.second, false);
+        w.second = true;
+
+        // Sanity check that a refresh would have been visible
+        w.refresh();
+        assert_eq!(w.second, false);
+    }
 }
--- a/src/write.rs
+++ b/src/write.rs
@@ -336,38 +336,12 @@ where
         self
     }
 
-    /// Gives the sequence of operations that have not yet been applied.
-    ///
-    /// Note that until the *first* call to `refresh`, the sequence of operations is always empty.
-    ///
-    /// ```
-    /// # use evmap::Operation;
-    /// let x = ('x', 42);
-    ///
-    /// let (r, mut w) = evmap::new();
-    ///
-    /// // before the first refresh, no oplog is kept
-    /// w.refresh();
-    ///
-    /// assert_eq!(w.pending(), &[]);
-    /// w.insert(x.0, x);
-    /// assert_eq!(w.pending(), &[Operation::Add(x.0, x)]);
-    /// w.refresh();
-    /// w.remove(x.0, x);
-    /// assert_eq!(w.pending(), &[Operation::Remove(x.0, x)]);
-    /// w.refresh();
-    /// assert_eq!(w.pending(), &[]);
-    /// ```
-    pub fn pending(&self) -> &[Operation<K, V>] {
-        &self.oplog[self.swap_index..]
-    }
-
     /// Refresh as necessary to ensure that all operations are visible to readers.
     ///
     /// `WriteHandle::refresh` will *always* wait for old readers to depart and swap the maps.
     /// This method will only do so if there are pending operations.
     pub fn flush(&mut self) -> &mut Self {
-        if !self.pending().is_empty() {
+        if !self.oplog[self.swap_index..].is_empty() {
             self.refresh();
         }
 
@@ -440,7 +414,7 @@ where
     ///
     /// The updated value-bag will only be visible to readers after the next call to `refresh()`.
     pub fn remove_value(&mut self, k: K, v: V) -> &mut Self {
-        self.add_op(Operation::Remove(k, v))
+        self.add_op(Operation::RemoveValue(k, v))
     }
 
     /// Remove the value-bag for the given key.
@@ -455,7 +429,7 @@ where
     ///
     /// The value-bag will only disappear from readers after the next call to `refresh()`.
     pub fn remove_entry(&mut self, k: K) -> &mut Self {
-        self.add_op(Operation::Empty(k))
+        self.add_op(Operation::RemoveEntry(k))
     }
 
     /// Purge all value-bags from the map.
@@ -593,7 +567,7 @@ where
                     .or_insert_with(Values::new)
                     .push(unsafe { value.shallow_copy() }, hasher);
             }
-            Operation::Empty(ref key) => {
+            Operation::RemoveEntry(ref key) => {
                 #[cfg(not(feature = "indexed"))]
                 inner.data.remove(key);
                 #[cfg(feature = "indexed")]
@@ -608,7 +582,7 @@ where
                     inner.data.swap_remove_index(index);
                 }
             }
-            Operation::Remove(ref key, ref value) => {
+            Operation::RemoveValue(ref key, ref value) => {
                 if let Some(e) = inner.data.get_mut(key) {
                     // remove a matching value from the value set
                     // safety: this is fine
@@ -671,7 +645,7 @@ where
                     .or_insert_with(Values::new)
                     .push(value, hasher);
             }
-            Operation::Empty(key) => {
+            Operation::RemoveEntry(key) => {
                 #[cfg(not(feature = "indexed"))]
                 inner.data.remove(&key);
                 #[cfg(feature = "indexed")]
@@ -686,7 +660,7 @@ where
                     inner.data.swap_remove_index(index);
                 }
             }
-            Operation::Remove(key, value) => {
+            Operation::RemoveValue(key, value) => {
                 if let Some(e) = inner.data.get_mut(&key) {
                     // find the first entry that matches all fields
                     e.swap_remove(&value);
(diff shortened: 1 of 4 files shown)


src/write.rs
```diff
@@ -431,14 +431,30 @@ where
     /// Remove the given value from the value-bag of the given key.
     ///
     /// The updated value-bag will only be visible to readers after the next call to `refresh()`.
+    #[deprecated(since = "11.0.0", note = "Renamed to remove_value")]
```
mlodato517, line 434: Is this the right version? I imagine the alpha you're on is therefore pre-`11.0.0`? In which case ... do we need to backwards compatible? I imagine it's a major release so ⚖️. Unless the current `11.0.0-alpha` is meant to be between `11.0.0` and `11.1.0` or some such?

src/write.rs
```diff
@@ -431,14 +431,30 @@ where
...
     pub fn remove(&mut self, k: K, v: V) -> &mut Self {
+        self.remove_value(k, v)
```
mlodato517, line 436: Is this the right way to deprecate this? So that if people are using the old one, they get new functionality? Not sure if that's good or not. Also not sure if this function should have its documentation removed so it doesn't get out of sync if `remove_value` changes? Or maybe I just shouldn't change the innards of `remove`?

src/write.rs
```diff
@@ -431,14 +431,30 @@ where
...
+    }
+
+    /// Remove the given value from the value-bag of the given key.
+    ///
+    /// The updated value-bag will only be visible to readers after the next call to `refresh()`.
+    pub fn remove_value(&mut self, k: K, v: V) -> &mut Self {
         self.add_op(Operation::Remove(k, v))
```
mlodato517, line 443: Should I also rename these operations? I imagine those are internal, implementation details and should be safe to rename? Although maybe not valuable ⚖️

tests/quick.rs
```diff
@@ -110,16 +110,16 @@ fn do_ops<K, V, S>(
...
             }
             RemoveValue(ref k, ref v) => {
-                evmap.remove(k.clone(), v.clone());
+                evmap.remove_value(k.clone(), v.clone());
                 write_ref.get_mut(k).and_then(|values| {
                     values
                         .iter_mut()
                         .position(|value| value == v)
-                        .and_then(|pos| Some(values.remove(pos)))
+                        .and_then(|pos| Some(values.swap_remove(pos)))
```
mlodato517, line 122: This is unrelated to this PR and is just cleanup from [this](https://github.com/jonhoo/rust-evmap/pull/58#issuecomment-654859029).

src/write.rs
```diff
@@ -431,14 +431,30 @@ where
...
     /// The value-bag will only disappear from readers after the next call to `refresh()`.
+    #[deprecated(since = "11.0.0", note = "Renamed to remove_entry")]
     pub fn empty(&mut self, k: K) -> &mut Self {
+        self.remove_entry(k)
+    }
+
+    /// Remove the value-bag for the given key.
+    ///
+    /// The value-bag will only disappear from readers after the next call to `refresh()`.
+    pub fn remove_entry(&mut self, k: K) -> &mut Self {
```
mlodato517, line 457: Oh, also, I later discovered that [`remove_entry`](https://doc.rust-lang.org/beta/std/collections/struct.HashMap.html#method.remove_entry) also exists for `std::collections::HashMap` so maybe we don't want to use this because this doesn't use the Entry API (yet?). Maybe this should be `remove_key` or `remove_bag` or something?

mlodato517:
Oh, and that code coverage is a good point - I imagine we should still be testing deprecated methods?

src/write.rs
jonhoo: You're totally right that we don't _have_ to, since we're about to make a breaking release anyway. However, I think having one major version with the methods deprecated is worthwhile. Makes transitioning a little nicer :)

src/write.rs
jonhoo: There's no new functionality here though, right? Since it's just a rename, I think it's fine.

mlodato517: Not yet! I was just thinking that if someone goes and updates `remove_value` for some reason, should those still using `remove` get that new change?

src/write.rs
jonhoo: It was precisely because of `HashMap::remove_entry` that I suggested the name :p The fact that it includes the key as well gives more of the flavor of "this removes the whole thing".

src/write.rs
jonhoo: Yes, I think renaming them is probably worthwhile.
I forget whether `Operation` is `pub` actually.. :/

jonhoo:
Hmm... No, I think it's fine to leave them out.

src/write.rs
mlodato517: Haha okay! Yeah I don't think we need more words to describe "removes the whole thing" (as the expected API is probably just "remove") but I didn't know if we should reach for "not `_entry`" because one might expect `remove_entry` should return `Option<(K, V)>` and this doesn't.

src/write.rs
jonhoo: If `remove_value` gets updated without its semantics change, then yes, I think `remove` should also get that change. And changing the semantics of `remove_value` would be backwards-incompatbile anyway. So I think it's fine.

src/write.rs
jonhoo: Nah, I think it's probably fine the way it is. People already realize pretty quickly that `evmap` doesn't have the same interface as `HashMap`.

src/write.rs
mlodato517: Yeah it looks like `Operation` is `pub`. So I imagine we want to deprecate them instead of a straight rename? That means that we have to match on deprecated enum variants in a few places. Is that okay? Or do you want me to just delete those? And do I need to do any sort of "cargo ignore this match of a deprecated enum" so it's not super annoying?

jonhoo: I think what I want to do is make `Operation` not be `pub`. That is a breaking change I think I'd be willing to make for `11.0`. If you change it to `pub(crate)`, how many things break?

mlodato517: It looks like maybe just [`pending`](https://github.com/jonhoo/rust-evmap/blob/master/src/write.rs#L361)?

jonhoo: Yeah, let's remove `pending` too. That's not really an API people should be using anyway.

mlodato517: This is a dream come true. Love deleting code.

mlodato517: Oh wait, you just meant make not public, I assume? Poor me 😢

mlodato517: So for [this test](https://github.com/jonhoo/rust-evmap/blob/master/tests/lib.rs#L188) do I remove the assertion and just put a dummy `assert!(true)` after the `flush()` or do I move the test into `src/write`?

jonhoo: I think you can actually just delete it. I don't think we use it internally for anything?
For that test, I'd move it to be a unit test, and then it can inspect the internal state that `pending()` accesses directly.

mlodato517: So dfb4843184f9c8aa3e974e8607f753138ef5d53a makes `Operation` `pub(crate)`, removes `pending()`, and moves the test

src/write.rs
```diff
@@ -336,38 +336,12 @@ where
...
-        &self.oplog[self.swap_index..]
-    }
-
     /// Refresh as necessary to ensure that all operations are visible to readers.
     ///
     /// `WriteHandle::refresh` will *always* wait for old readers to depart and swap the maps.
     /// This method will only do so if there are pending operations.
     pub fn flush(&mut self) -> &mut Self {
-        if !self.pending().is_empty() {
+        if !self.oplog[self.swap_index..].is_empty() {
```
mlodato517, line 344: Now that we don't have `pending()` for the added readability, I kind of want to just check
```suggestion
        if self.swap_index < self.oplog.len() {
```
or whatever makes this work 🤔

jonhoo: Yes, that should be right. Running the test suite should confirm it pretty quickly.

mlodato517: I like the confidence! I wasn't sure because it looked like [only one test called `flush` and it didn't test much](https://github.com/jonhoo/rust-evmap/blob/master/tests/lib.rs#L189) but I'll go ahead with that change

mlodato517: I guess I should maybe update/write some test to ensure that `refresh` isn't called if I call `flush` twice or something? Is that a thing that is possible to verify?

jonhoo: Huh, interesting. I thought `flush` was called internally, but I guess not. Yes, then adding a test that checks that would be useful indeed! You'll probably need to write it as a unit test and then inspect some internal state to check if `refresh` was called.

mlodato517: Added 351881d37f68326de12618ff8606969e76d1a876. I'm not sure if `w.second` is the best idea but there wasn't anything _terribly_ obvious

src/write.rs
```diff
@@ -748,4 +748,46 @@ mod tests {
...
+    fn flush_no_refresh() {
+        let x = 'x';
+
+        let (_, mut w) = new();
+
+        // Until we refresh, writes are written directly instead of going to the
+        // oplog (because there can't be any readers on the w_handle table).
+        w.refresh();
+
+        assert_eq!(w.second, true);
```
jonhoo, line 762: Maybe assert `!w.second` before the first `refresh`?

src/write.rs
```diff
@@ -748,4 +748,46 @@ mod tests {
...
+
+        w.insert(x, x);
+        w.flush();
+
+        // A refresh happened!
+        assert_eq!(w.second, false);
+
+        // Reset this flag so we can see if another refresh happens
+        w.second = true;
```
jonhoo, line 776: I wonder what kind of weird, deeper effects changing this externally might have... Hopefully none :sweat_smile:

mlodato517: Haha yeah this feels ... horrible ... but I'm not sure how else to check, "Did a refresh happen"? Is there some other flag/meta data I could check?

jonhoo: To be honest, I'd probably just add a:
```rust
#[cfg(test)]
refreshes: usize
```
field to `WriteHandle`, and then have a
```rust
#[cfg(test)]
self.refreshes += 1;
```
in the appropriate place in `refresh`. That way, in testing code, you'll know exactly how many refreshes have happened. And it wouldn't require hackery or impact running code.

mlodato517: Hmmm maybe I just have the wrong syntax but I'm getting some
```
attributes on expressions are experimental
removing an expression is not supported in this position
```
when I try
```rust
pub struct WriteHandle {
    // ...
    #[cfg(test)]
    refreshes: usize,
}
pub(crate) fn new(/*...*/) -> WriteHandle {
    let m = w_handle.meta.clone();
    WriteHandle {
        // ...
        #[cfg(test)]
        refreshes: 0,
    }

  pub fn refresh(&mut self) -> &mut Self {
    // ...
    #[cfg(test)]
    self.refreshes += 1;
```
Was that what you had in mind?

jonhoo: Ugh, I forgot they're still experimental. Just do this instead:
```rust
#[cfg(test)]
{
  self.refreshes += 1;
}
```
and that should work I think

mlodato517: I'm a little wigged out about sprinkling the test cfg around too much but https://github.com/jonhoo/rust-evmap/pull/60 might do it

Issue #17, opened 2019-02-22
Test with loom


jonhoo:
@carllerche recently published [`loom`](https://docs.rs/loom/) which does correctness checking for concurrent algorithms. It'd be neat to try to get that working for `evmap`'s tests!

jonhoo:
cc @novacrazy

aDotInTheVoid:
I've started work on this [here](https://github.com/aDotInTheVoid/rust-evmap/tree/loom)

aDotInTheVoid:
Currently, this is blocked on loom [only supporting](https://github.com/tokio-rs/loom/blob/2e030653e78d83a8fb46d808397e693494e3bdcc/src/rt/atomic.rs#L142-L145) Aquire fences.

jonhoo:
Hmm, my _guess_ here is that you can just test with `Acquire` fences, and it should be fine. Or at least you should be able to make progress. You can see what I did to deal with this in crossbeam: https://github.com/crossbeam-rs/crossbeam/blob/03e04ba4426e47f6b76d9e9da551c81157f9cbb0/crossbeam-epoch/src/lib.rs#L81-L92

jonhoo:
@aDotInTheVoid Did you have a change to try out using a stronger fence for testing?

jonhoo:
Now that `evmap` is split up into `left-right` and `evmap`, it _should_ be easier to at least get `loom` testing for `left-right` itself!

djc:
Looks like this is done?

PR #61, opened 2020-07-09
Adds keys and values iterators

mlodato517:
**This PR**
Adds `keys` and `values` iterator helpers to the `MapReadRef` per [this comment](https://github.com/jonhoo/rust-evmap/pull/58#discussion_r450439632).

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/rust-evmap/61)

### DIFF
--- a/src/read/read_ref.rs
+++ b/src/read/read_ref.rs
@@ -43,16 +43,20 @@ where
     ///
     /// Be careful with this function! While the iteration is ongoing, any writer that tries to
     /// refresh will block waiting on this reader to finish.
-    pub fn keys(&self) -> impl Iterator<Item = &K> {
-        self.iter().map(|(k, _)| k)
+    pub fn keys(&self) -> KeysIter<'_, K, V, S> {
+        KeysIter {
+            iter: self.guard.data.iter(),
+        }
     }
 
-    /// Iterate over all values in the map.
+    /// Iterate over all value sets in the map.
     ///
     /// Be careful with this function! While the iteration is ongoing, any writer that tries to
     /// refresh will block waiting on this reader to finish.
-    pub fn values(&self) -> impl Iterator<Item = &V> {
-        self.iter().map(|(_, v)| v).flatten()
+    pub fn values(&self) -> ValuesIter<'_, K, V, S> {
+        ValuesIter {
+            iter: self.guard.data.iter(),
+        }
     }
 
     /// Returns the number of non-empty keys present in the map.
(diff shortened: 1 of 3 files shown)


tests/lib.rs
```diff
@@ -593,6 +593,48 @@ fn map_into() {
...
+#[test]
+fn keys() {
+    let (r, mut w) = evmap::new();
+    w.insert(1, "a");
+    w.insert(1, "b");
+    w.insert(2, "c");
+    w.refresh();
+    w.insert(1, "x");
+
+    let mut keys = r
```
mlodato517, line 605: Looking at this usage ... this doesn't look very helpful ...
I wonder if it would be better to have `r.keys_blocking()` or something so that the user doesn't have to deal with flattening out the `Some()`s and is fairly aware of the blocking nature?

src/read/read_ref.rs
```diff
@@ -39,6 +39,22 @@ where
         }
     }
 
+    /// Iterate over all keys in the map.
```
mlodato517, line 42: One thing I found interesting ... which is just off screen, is that `iter` is `Some(self.guard.data.iter())` ... why is it optional? Is it ever `None`?

src/read/read_ref.rs
```diff
@@ -39,6 +39,22 @@ where
...
+    ///
+    /// Be careful with this function! While the iteration is ongoing, any writer that tries to
+    /// refresh will block waiting on this reader to finish.
+    pub fn keys(&self) -> impl Iterator<Item = &K> {
```
mlodato517, line 46: Should I be using `impl Iterator` here or should I make a new `Keys` struct? Ditto for down below.

jonhoo: So, the thing that's nice about an explicit type is that you can also implement `ExactSizeIterator`, `DoubleEndedIterator`, and `FusedIterator`. You can't do that (yet) with `impl Trait`.

src/read/read_ref.rs
jonhoo: Hmm, that might be leftover from before we had `read()` and `MapReadRef`. If the `Option` can be removed now, then please do!

tests/lib.rs
jonhoo: I'd really rather avoid that. The use-case for `evmap` is generally that you're _first_ going to get access to the read ref (through `read()`), and only if that succeeds do you do the operation. I'd like to not shorthand going through `read`, because it'd basically mean we should duplicate all the methods.

src/read/read_ref.rs
```diff
@@ -39,6 +39,22 @@ where
...
+        self.iter().map(|(k, _)| k)
+    }
+
+    /// Iterate over all values in the map.
+    ///
+    /// Be careful with this function! While the iteration is ongoing, any writer that tries to
+    /// refresh will block waiting on this reader to finish.
+    pub fn values(&self) -> impl Iterator<Item = &V> {
```
mlodato517, line 54: One more question - should `values()` return an iterator over `&V` or over `Values<V, ...>`. That is, should be a flat iterator over each value or should it be an iterator over each value set? If this were `HashMap<K, Vec<V>>` then I'd expect over each value set but since you don't get the choice of not having the value collection I didn't know what the "expected" value type is.

jonhoo: I think it should be an iterator over the `Values<V>`. The user can always choose to flatten with `Iterator::flatten()`.

src/read/read_ref.rs
mlodato517: c79d3c027e68c49346bcd2178cbd7708836383b2

src/read/read_ref.rs
mlodato517: 61c6642bb77fb25c2428222a2c12bd1dabaf8dc7

jonhoo:
This looks great once again, thank you!

mlodato517:
You're welcome! And thanks for all your great reviews!

While writing [these assertions](https://github.com/jonhoo/rust-evmap/pull/61/files#diff-934f6ed12102bba0b289ee2855cb9bd1R243) I was thinking, "It might be convenient to implement `Eq` on `Values` if the inner value implements `Eq`". But I can't tell if it's just convenient for tests. Is that something you might want?

Otherwise, looking through [the `indexmap` tests](https://github.com/bluss/indexmap/blob/master/tests/quick.rs) there are tests for `map.values_mut()`, `map1 == map2`, and then some stuff based on sort order that I'm guessing we don't care about because we don't care about order. Would you prefer that kind of functionality?

Above and beyond all that, are there any other things you've been looking to revisit/clean up that I could tackle? [This comment](https://github.com/jonhoo/rust-evmap/blob/master/src/write.rs#L260) looks like an obvious place to start. I could think on that if it's valuable. I'm not sure I'd be able to brain out something at this level, but I'd also be interested in trying 🤷

Or if there are any other projects you have or know about that need some more chores like this done (or easy features or whatever), I'm game!

jonhoo:
It's been a pleasure!

I think `PartialEq` + `Eq` on `Values` is _probably_ mostly useful for tests, but it seems reasonable enough to me! One thing that'll be tricky is to check that the cardinality is right if one thing is `Short` (but has duplicates) and the other thing is a `Long`.

I think the particular code comment you linked to is unlikely to matter much in practice, since it only every runs once for a given `evmap`. In general, a good strategy is to look for things that are missing in the [coverage map](https://codecov.io/gh/jonhoo/rust-evmap/tree/master/src) and add tests for those. In terms of other issues, it would be _awesome_ to add loom tests (https://github.com/jonhoo/rust-evmap/issues/17), which will also require some thinking to get right.

As far as other projects go, here are a couple:
 - `griddle` has some chore-like stuff in https://github.com/jonhoo/griddle/issues/4 (`RawEntry` in particular)
 - `inferno` has some test coverage bits in https://github.com/jonhoo/inferno/issues/63
 - `flurry` needs more tests here https://github.com/jonhoo/flurry/issues/20

PR #62, opened 2020-07-17
[Draft] Add Loom Tests

mlodato517:
Opening up a draft PR to start some discussion!

Most of the early questions I have are things like, "`loom`'s `Mutex` doesn't implement `Default` which causes a bunch of compilation failures. Should we PR that into their library first or do they not do `Default` for a good reason and we have to come up with a workaround?"

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/rust-evmap/62)

### DIFF
--- a/src/read/mod.rs
+++ b/src/read/mod.rs
@@ -163,7 +163,7 @@ where
 
         // so, update our epoch tracker.
         let epoch = self.my_epoch.fetch_add(1, Ordering::Relaxed);
-        self.epoch.store(epoch + 1, Ordering::Release);
+        // self.epoch.store(epoch + 1, Ordering::Release);
 
         // ensure that the pointer read happens strictly after updating the epoch
         fence(Ordering::SeqCst);
--- a/src/read/mod.rs
+++ b/src/read/mod.rs
@@ -1,15 +1,15 @@
 use crate::inner::Inner;
 use crate::values::Values;
 
+use crate::sync::{fence, Arc, AtomicPtr, AtomicUsize};
 use std::borrow::Borrow;
 use std::collections::hash_map::RandomState;
 use std::hash::{BuildHasher, Hash};
 use std::iter::FromIterator;
 use std::marker::PhantomData;
 use std::mem::ManuallyDrop;
-use std::sync::atomic;
-use std::sync::atomic::AtomicPtr;
-use std::sync::{self, Arc};
+// REMOVE THIS BEFORE REVIEWING
+use std::sync::atomic::Ordering;
 use std::{cell, fmt, mem};
 
 mod guard;
@@ -31,11 +31,11 @@ where
     K: Eq + Hash,
     S: BuildHasher,
 {
-    pub(crate) inner: sync::Arc<AtomicPtr<Inner<K, ManuallyDrop<V>, M, S>>>,
+    pub(crate) inner: Arc<AtomicPtr<Inner<K, ManuallyDrop<V>, M, S>>>,
     pub(crate) epochs: crate::Epochs,
-    epoch: sync::Arc<sync::atomic::AtomicUsize>,
+    epoch: Arc<AtomicUsize>,
     epoch_i: usize,
-    my_epoch: sync::atomic::AtomicUsize,
+    my_epoch: AtomicUsize,
 
     // Since a `ReadHandle` keeps track of its own epoch, it is not safe for multiple threads to
     // call `with_handle` at the same time. We *could* keep it `Sync` and make `with_handle`
@@ -79,10 +79,7 @@ where
     S: BuildHasher,
 {
     fn clone(&self) -> Self {
-        ReadHandle::new(
-            sync::Arc::clone(&self.inner),
-            sync::Arc::clone(&self.epochs),
-        )
+        ReadHandle::new(Arc::clone(&self.inner), Arc::clone(&self.epochs))
     }
 }
 
@@ -95,7 +92,7 @@ where
     S: BuildHasher,
 {
     let store = Box::into_raw(Box::new(inner));
-    ReadHandle::new(sync::Arc::new(AtomicPtr::new(store)), epochs)
+    ReadHandle::new(Arc::new(AtomicPtr::new(store)), epochs)
 }
 
 impl<K, V, M, S> ReadHandle<K, V, M, S>
@@ -103,12 +100,9 @@ where
     K: Eq + Hash,
     S: BuildHasher,
 {
-    fn new(
-        inner: sync::Arc<AtomicPtr<Inner<K, ManuallyDrop<V>, M, S>>>,
-        epochs: crate::Epochs,
-    ) -> Self {
+    fn new(inner: Arc<AtomicPtr<Inner<K, ManuallyDrop<V>, M, S>>>, epochs: crate::Epochs) -> Self {
         // tell writer about our epoch tracker
-        let epoch = sync::Arc::new(atomic::AtomicUsize::new(0));
+        let epoch = Arc::new(AtomicUsize::new(0));
         // okay to lock, since we're not holding up the epoch
         let epoch_i = epochs.lock().unwrap().insert(Arc::clone(&epoch));
 
@@ -116,7 +110,7 @@ where
             epochs,
             epoch,
             epoch_i,
-            my_epoch: atomic::AtomicUsize::new(0),
+            my_epoch: AtomicUsize::new(0),
             inner,
             _not_sync_no_feature: PhantomData,
         }
@@ -126,8 +120,8 @@ where
     /// threads.
     pub fn factory(&self) -> ReadHandleFactory<K, V, M, S> {
         ReadHandleFactory {
-            inner: sync::Arc::clone(&self.inner),
-            epochs: sync::Arc::clone(&self.epochs),
+            inner: Arc::clone(&self.inner),
+            epochs: Arc::clone(&self.epochs),
         }
     }
 }
@@ -168,14 +162,14 @@ where
         // in all cases, using a pointer we read *after* updating our epoch is safe.
 
         // so, update our epoch tracker.
-        let epoch = self.my_epoch.fetch_add(1, atomic::Ordering::Relaxed);
-        self.epoch.store(epoch + 1, atomic::Ordering::Release);
+        let epoch = self.my_epoch.fetch_add(1, Ordering::Relaxed);
+        self.epoch.store(epoch + 1, Ordering::Release);
 
         // ensure that the pointer read happens strictly after updating the epoch
-        atomic::fence(atomic::Ordering::SeqCst);
+        fence(Ordering::SeqCst);
 
         // then, atomically read pointer, and use the map being pointed to
-        let r_handle = self.inner.load(atomic::Ordering::Acquire);
+        let r_handle = self.inner.load(Ordering::Acquire);
 
         // since we bumped our epoch, this pointer will remain valid until we bump it again
         let r_handle = unsafe { r_handle.as_ref() };
@@ -191,7 +185,7 @@ where
             // the map has not yet been initialized, so restore parity and return None
             self.epoch.store(
                 (epoch + 1) | 1usize << (mem::size_of::<usize>() * 8 - 1),
-                atomic::Ordering::Release,
+                Ordering::Release,
             );
             None
         }
@@ -201,7 +195,7 @@ where
     /// the returned reference.
     pub(crate) unsafe fn hasher(&self) -> &S {
         self.inner
-            .load(atomic::Ordering::Acquire)
+            .load(Ordering::Acquire)
             .as_ref()
             .unwrap()
             .data
--- a/tests/loom.rs
+++ b/tests/loom.rs
@@ -1,0 +1,32 @@
+#[cfg(loom)]
+#[cfg(test)]
+mod loom_tests {
+    use loom::sync::atomic::AtomicUsize;
+    use loom::sync::atomic::Ordering::{Acquire, Relaxed, Release};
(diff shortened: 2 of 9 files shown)


jonhoo:
I think we should prefer submitting PRs to loom where the fixes are simple. Better to improve the ecosystem than have our own hacks! @carllerche may also be able to provide some pointers if you have specific questions. I'd probably ask over on the Tokio Discord, which has a #loom channel :)

mlodato517:
Hmm so actually it looks like it [has been added](https://github.com/tokio-rs/loom/commit/871ff0128888dea307b848549dd0a5692b1e82ad) it's just not in any formally released version yet 🤔 Should I point the dependency at their `master` branch?

jonhoo:
Sure, yeah, do that for the purposes of moving forward with the PR, and then we'll just wait for it to land before we actually merge!

mlodato517:
Okay so my original plan was going to be:
1. Add `loom`
1. Mark every existing test with `#[cfg(not(loom))]`
1. Identify which tests we wanted to model with `loom`/write new tests with `loom` that we want

Question 1 - Do we want to have every test wrapped in `loom` by default or is the second step the way we want to go?
Question 2 - How should I handle doc tests? I imagine any `loom` stuff would clutter the documentation but I imagine there's a way to hide it. I tried:
```rust
//! # #[cfg(not(loom))]
//! # {
//! let (book_reviews_r, mut book_reviews_w) = evmap::new();
//! // ...
//! # }
```
but that didn't work 🤔 8cbe2e5dde799e8ea6ccef73030b3b424b98d841

jonhoo:
1. The common way to do this is to mark the whole test module as `cfg(not(loom))`. For unit tests, this would be:
   ```rust
   #[cfg(not(loom))]
   mod tests {
   ```
   (like you already have) and for integration tests you'd add
   ```rust
   #![cfg(not(loom))]
   ```
   to the top of the file (this latter trick I think is the one you're missing).
2. We should just have CI not run loom on doctests. The easiest way to do this is to have the invocation command be:
   ```console
   $ env RUSTFLAGS="--cfg loom" cargo test --lib --tests
   ```
   We may even be able to avoid the `#[cfg(not(loom))]` on unit tests by dropping the `--lib`, but you'd have to check that that actually works first.

mlodato517:
>        (like you already have) and for integration tests you'd add
>        ```rust
>        #![cfg(not(loom))]
>        ```
>        
>        to the top of the file (this latter trick I think is the one you're missing).

Ohh! So does `!` here act the same as `//!` and like "apply to the anonymous top level module" or something? If so, TIL!

>     2. We should just have CI not run loom on doctests. The easiest way to do this is to have the invocation command be:
>        ```
>        $ env RUSTFLAGS="--cfg loom" cargo test --lib --tests
>        ```

Oh hmmm yeah I was thinking of, similar to `tests/quick` having `tests/loom`. Would we then just have to do the loom configuration for that one file? Especially if we're going the route of "tests are not loom configurable by default". Or is that not possible? And, if that's the way we want to go, is it worth documenting anywhere that "when running loom tests, only target xyz"?

jonhoo:
> Ohh! So does `!` here act the same as `//!` and like "apply to the anonymous top level module" or something? If so, TIL!

Exactly!

>  Would we then just have to do the loom configuration for that one file?

Ah, that's a good point. You may actually be able to get rid of all those `cfg(not(loom))` and instead just run loom with
```console
$ env RUSTFLAGS="--cfg loom" cargo test --test loom
```

mlodato517:
So my brain's having a small issue of what tests to write. The only thing I can really think of is "Start a reader on one/x thread and a writer on another thread. Write some stuff. Refresh. Ensure reader(s) on other thread read correct value". But I'm not sure if that's useful/powerful enough. I re-watched the video about Noria and feel like there's another test of like, "When the refresh happens, new writes won't accidentally interfere with writers on the 'old' table". I kind of feel like we want to test this ... but I don't know how.

Am I heading in the right direction with this at all?

mlodato517:
Also, sorry, I've never used Azure before. I'm looking through the [pipelines file](https://github.com/jonhoo/rust-evmap/blob/master/azure-pipelines.yml) and trying to figure out how to add the step for loom tests On the [pipeline page (or whatever this is)](https://dev.azure.com/jonhoo/jonhoo/_build/results?buildId=1635&view=logs&jobId=1dca0c1e-db73-5ce7-56ee-f8fcc3db08e5) I see all sorts of stuff for "Compile and test Linux (nightly)", "Compile and test Linux (beta)", "Compile and test MacOS", etc. but I don't see those in the pipelines file. Is it because those are part of a template or am I just reading it wrong?

I assumed we wanted to add the loom tests to all the OSs and versions and things but I didn't want to add like 7 new jobs. I figured I'd either want to add it to the templates or somehow make it part of `cargo test` or something witchcrafty?

jonhoo:
Start simple — once the infrastructure is in place, we can always add more tests later :)
The kinds of tests I imagine we'll want are things like having a reader and a writer race to delete a key from the map, to check that the deallocation only happens when it is truly safe, or the writer overwriting a value while the reader holds it, and make sure things still work.

As for CI, you're entirely right that it comes from a template. Specifically [this one](https://github.com/crate-ci/azure-pipelines/blob/master/default.yml). I think loom definitely should be in its own job, so don't worry too much about adding it separately. I also don't think we need to run the loom tests on anything but Linux, that should be fine. Only running it on one Rust version should also be totally acceptable. All that matter is catching the races, and if they are there they should be independent of both those things.

mlodato517:
> I think loom definitely should be in its own job, so don't worry too much about adding it separately. I also don't think we need to run the loom tests on anything but Linux, that should be fine. Only running it on one Rust version should also be totally acceptable.

👍 

> having a reader and a writer race to delete a key from the map, to check that the deallocation only happens when it is truly safe

I'm a little confused here. I can imagine various things like:

1. We read the value on one thread, delete it on the other thread, join the threads, and verify the reader saw the value
1. We read the value on one thread, delete it on the other thread, join the threads, refresh, read again(?), and verify the reader sees no value
1. ~We read the value on one thread, delete + refresh on the other thread, join the threads ... no wait this one isn't deterministic because the it depends on if the reader read before or after the refresh~
1. We read the value on one thread, refresh + delete it on the other thread, join the threads, and verify that the reader saw the value

Is that about what you're thinking? I can't tell if these are "useful" because until `refresh`es happen there isn't any synchronization logic and after a refresh happens we have to re-read the map to see anything? But it could just be too Monday for me!

jonhoo:
The way to think about this is that `loom` tries to test for whether our expectations are actually what happen in practice. You're right that _in theory_ nothing should happen until `refresh` is called, and `refresh` should be atomic, and so these tests should be "useless". But if there's a big, and they aren't, then the tests would fail. And so they aren't useless. For example, a loom test that tries modifying a map immediately after calling `refresh` might catch a case where `refresh` does not correctly wait for all readers to depart an old map before returning, which would lead to a race.

src/read/mod.rs
```diff
@@ -168,14 +162,14 @@ where
         // in all cases, using a pointer we read *after* updating our epoch is safe.
 
         // so, update our epoch tracker.
-        let epoch = self.my_epoch.fetch_add(1, atomic::Ordering::Relaxed);
-        self.epoch.store(epoch + 1, atomic::Ordering::Release);
+        let epoch = self.my_epoch.fetch_add(1, Ordering::Relaxed);
+        // self.epoch.store(epoch + 1, Ordering::Release);
```
mlodato517, line 166: This is pretty interesting. This logic just might be too complicated for the current `loom` capabilities. If I uncomment this line and run the test in `tests/loom` (which I _think_ is right and I _think_ is about as simple as it's going to get?) I get:
> thread 'loom_tests::evmap_read_while_remove' panicked at 'Model exeeded maximum number of branches. This is often caused by an algorithm requiring the processor to make progress,
e.g. spin locks.'

If I comment this out, the test passes. We might need to either simplify the logic or start getting some PRs into `loom` to beef it up? Currently it looks like the thread max is handled by a `u16` and maybe we can just bump that up? 🤷 Interested in your thoughts! Would also be free sometime to jump on a call to work through this in a more synchronous manner depending upon both our lives

mlodato517: And obviously this line isn’t special - just showing that we might be at the cusp of complexity

mlodato517: Or maybe I just did a bad job increasing the default max with env vars

tests/loom.rs
```diff
@@ -0,0 +1,36 @@
...
+
+            let val = Arc::new(AtomicUsize::new(0));
+            let val_copy = Arc::clone(&val);
+
+            let read_thread = thread::spawn(move || {
+                val.store(*r.get_one(&1).as_deref().unwrap(), SeqCst);
+            });
+
+            let write_thread = thread::spawn(move || {
+                w.remove_entry(1);
```
jonhoo, line 27: You don't need to spawn a thread for performing this operation. You can just perform it directly on the main thread, and loom will still handle scheduling the different interleavings.

src/read/mod.rs
jonhoo: Hmm, that's interesting. We _are_ spinning in the writer waiting for the reader to finish its read, that is certainly true. I think maybe the trick here is going to be to have
https://github.com/jonhoo/rust-evmap/blob/34b5306be23063579dcf5a7aefba93af39eb4ade/src/write.rs#L204

Always call `yield_now` rather than spin when running under `loom`. That might fix it.

mlodato517: I was wondering about this because we don't call `refresh` after the reader so I didn't think this was possible. I was only looking at `remove_entry` which didn't do anything interesting.

Is it that `loom` has scheduled the `refresh` at the beginning of the test to be after when the reader reads? I assumed it wouldn't move around "sequential" operations (which is why I also had the `w.remove_entry` in a thread - do I just need no threads and it'll reorder everything anyway?). Maybe I need to read their docs more slowly and/or do a mini deep dive into the code to figure out how it works.

tests/loom.rs
mlodato517: Oh so this is something I was messing around with yesterday when I was trying to figure out what it was doing. If I don't spawn a thread here the test just passes instantly.

src/read/mod.rs
jonhoo: Hmm, no, you're right, since there isn't a refresh after the operation, that shouldn't be spinning, since as you say spawning a thread is a synchronizing operation. I still think we'll need the change I outlined above for other tests, but not for this one.

I'd try jumping in the loom channel and see if anyone can shed some light there.

tests/loom.rs
mlodato517: Also, and I don't know if I'm doing this right, if I remove this thread and then put an `eprintln` in `remove_entry` and `get_one` I only see this output in the test:
```
running 1 test
Removing entry!
Reading value!
Completed in 1 iterations
test loom_tests::evmap_read_while_remove ... ok
```
Which makes me think `loom` never tried the version where we read the value before removing the entry? Am I interpreting that correctly?

mlodato517: Unless `loom` determined through it's state-squashing magic that there was no need to reorder them 🤔

jonhoo: Ah, yes, this is because there's no synchronization happening on the write path here! Basically, the call to `remove_entry` just adds a remove to the operation log, and does nothing else, so loom (correctly) determines that there is no need to try another schedule. Running the other thread cannot possibly affect the execution of the remover thread. If you had a refresh in there on the other hand, it would. You could force it by using `thread::yield_now`, but basically loom eliminating this state is sufficient proof. If we _did_ add synchronization to `remove_entry` later, then loom _would_ run that additional execution and tell us if there was a mistake.

mlodato517: That makes sense 👍 

Follow up question: Is it worth writing `loom` tests for things where there are no "extra" scheduling options? Currently this is doing "nothing special" on top of the normal test suite. Do we want this here so that, if `remove_entry` _gains_ some synchronization code, we're "ready" for it?

I'm torn because I could see this catching an error if synchronization is added but I'm generally against writing code for futures which may never exist. This isn't my library and there may be some solid value there but my initial gut reaction is that the loom tests should be added when the synchronization logic is added. And this PR would only be adding those `loom` tests that actually exercise various thread sync logic.

What do you think?

jonhoo: My cautious vote would be to keep it. These things are subtle enough that guarding towards accidental future changes are worthwhile. Especially since this test takes so little time to run (as opposed to most loom tests). I would make sure to include a comment saying why it's there given its perceive lack of effect though.

jonhoo:
This should in theory be a _lot_ easier now by testing `left-right` directly rather than `evmap` (which has [moved](https://github.com/jonhoo/evmap)). You'll probably just test a `left_right::Handle<Box<usize>>` or something directly!

mlodato517:
> This should in theory be a _lot_ easier now by testing `left-right` directly rather than `evmap` (which has [moved](https://github.com/jonhoo/evmap)). You'll probably just test a `left_right::Handle<Box<usize>>` or something directly!

Awesome! So, just checking, is the idea is to:
1. close this PR and do something similar in `left-right`
2. close this PR and move it to the new `evmap` and write tests against `left_right::Handle...`
3. something else

?

jonhoo:
I think we should focus on `left-right`, since that's the reusable component. And because it'll probably be much easier to get off the ground with. So option 1 -- we should probably close this since it's getting pretty stale, and then start a new one that just adds loom support to `left-right` :)

mlodato517:
Oh shoot! I didn't notice that this repo _became_ `left-right`! Okay, yeah, I'll close this for that very reason and hopefully get this done ... soon

PR #64, opened 2020-07-21
Removes use of deprecated method in example

mlodato517:
When I was working on adding the `loom` tests (for https://github.com/jonhoo/rust-evmap/pull/62) I was seeing:
```
---- src/lib.rs -  (line 35) stdout ----
warning: use of deprecated item 'evmap::WriteHandle::<K, V, M, S>::empty': Renamed to remove_entry
  --> src/lib.rs:68:16
   |
36 | book_reviews_w.empty("The Adventures of Sherlock Holmes");
   |                ^^^^^
   |
   = note: `#[warn(deprecated)]` on by default
```

I don't see it normally when I `cargo check --examples` so I'd love to figure that out ... but in the meantime, this updates the one obvious example from #59

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/rust-evmap/64)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -63,8 +63,8 @@
 //! assert_eq!(book_reviews_r.get("Grimms' Fairy Tales").map(|rs| rs.len()), Some(2));
 //!
 //! // oops, this review has a lot of spelling mistakes, let's delete it.
-//! // empty deletes *all* reviews (though in this case, just one)
-//! book_reviews_w.empty("The Adventures of Sherlock Holmes");
+//! // remove_entry deletes *all* reviews (though in this case, just one)
+//! book_reviews_w.remove_entry("The Adventures of Sherlock Holmes");
 //! // but again, it's not visible to readers until we refresh
 //! assert_eq!(book_reviews_r.get("The Adventures of Sherlock Holmes").map(|rs| rs.len()), Some(1));
 //! book_reviews_w.refresh();


jonhoo:
Oh, yeah, doctests are never run by `cargo check` :)

mlodato517:
Ohh so we only see the deprecated warnings when the tests are run - not when cargo checks the examples?

jonhoo:
Ah, but this wasn't an example, this was a doctest. Which are only ever compiled when `cargo t` runs doctests!

mlodato517:
Ahh okay that makes more sense. But when I do `cargo test`, it runs doc tests and doesn't report anything. And `cargo test --doc` doesn't report anything ... I wonder why it showed up when I was doing the loom tests 🤔 Be nice if it showed up more readily!

jonhoo:
Oh, that's really weird then... Hmm.... Yeah, don't know then? Are you using a different compiler version or something when working with `loom`? Or maybe different feature flags?

mlodato517:
I think the only difference is `RUSTFLAGS="--cfg loom" cargo test --doc` vs `cargo test --doc`

jonhoo:
That's bizarre. I don't see why the former should surface more warnings than the latter.. Unless it's simply that the former had to re-compile, whereas the latter just re-ran the already-compiled doctest binaries? Could you try making a minor change and running `cargo test --doc` again?

Actually, also, can you try just bumping the minor version in `Cargo.toml` and then re-run `cargo test --doc`?

mlodato517:
It would appear that, when I switch between `RUSTFLAGS="--cfg loom" cargo test --doc` and regular `cargo test --doc` it recompiles anyway (maybe because of the different config?):
```
~/p/rust-evmap (loom-tests↑1|✔) $ RUSTFLAGS="--cfg loom" cargo test --doc
   Compiling cfg-if v0.1.10
   ...
     Finished test [unoptimized + debuginfo] target(s) in 10.45s
   Doc-tests evmap

running 7 tests
test src/lib.rs -  (line 177) ... ok
...

failures:

---- src/write.rs - write::WriteHandle (line 29) stdout ----
...

---- src/lib.rs -  (line 35) stdout ----
warning: use of deprecated item 'evmap::WriteHandle::<K, V, M, S>::empty': Renamed to remove_entry
...

~/p/rust-evmap (loom-tests↑1|✔) $ cargo test --doc
   Compiling libc v0.2.71
   ...
    Finished test [unoptimized + debuginfo] target(s) in 10.72s
   Doc-tests evmap

running 7 tests
test src/lib.rs -  (line 177) ... ok
...

test result: ok. 7 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

But, in any case, if I bump the version, I still don't see the warning without `RUSTFLAGS="--cfg loom"`. Although I also don't see it re-compiling ... I bumped the version of of the `evmap` package - is that what you meant? Or did you mean the minor version of some dependency? If I knock `rand` down to `0.6` I see the recompilation but no deprecation warning 🤷

jonhoo:
I have no idea then :shrug: Maybe file an issue on rust-lang/rust?

PR #66, opened 2020-08-10
Feature for amortized resizes

jonhoo:
Keep in mind that since this is a feature, _all_ evmap instances get amortized if _anything_ in the dependency graph enables the feature (https://github.com/bluss/indexmap/pull/137#issuecomment-662701746). This is unfortunate, but given that the alternative is forking evmap, we'll go with the feature approach for now. Hopefully one day we'll have a way to [avoid this](https://github.com/rust-lang/cargo/issues/2980#issuecomment-667207117).

Not landing this yet since it's nightly-only at the moment due to certain methods [in `atone`](https://github.com/jonhoo/atone/blob/45ae1e42deaaaaa9a919957ade49982a7ac4655b/src/lib.rs#L58). Nightly-only will go away once https://github.com/rust-lang/rust/issues/70929 and https://github.com/rust-lang/rust/issues/74217 both stabilize. The first one will land on next stable, but the second has some more time to go first.

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/rust-evmap/66)

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -7,11 +7,22 @@ stages:
         dir: "left-right"
  - stage: evmap
    jobs:
-    - template: default.yml@templates
+    - template: nightly-only.yml@templates
       parameters:
-        minrust: 1.40.0
         codecov_token: $(CODECOV_TOKEN_SECRET)
         dir: "evmap"
+    - job: features
+      displayName: "Check feature combinations"
+      pool:
+        vmImage: ubuntu-latest
+      steps:
+        - template: install-rust.yml@templates
+          parameters:
+            rust: nightly
+        - script: cargo install cargo-hack
+          displayName: install cargo-hack
+        - script: cargo hack --feature-powerset check
+          displayName: cargo hack
     - job: benchmark
       displayName: "Check that benchmark compiles"
       pool:
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -7,22 +7,11 @@ stages:
         dir: "left-right"
  - stage: evmap
    jobs:
-    - template: nightly-only.yml@templates
+    - template: default.yml@templates
       parameters:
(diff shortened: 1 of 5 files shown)


src/inner.rs
```diff
@@ -2,10 +2,12 @@ use std::fmt;
...
 use std::mem::ManuallyDrop;
 
-#[cfg(feature = "indexed")]
-pub(crate) use indexmap::IndexMap as MapImpl;
-#[cfg(not(feature = "indexed"))]
-pub(crate) use std::collections::HashMap as MapImpl;
+#[cfg(all(feature = "indexed", not(feature = "amortize")))]
+pub(crate) use indexmap::{map::Entry, IndexMap as MapImpl};
+#[cfg(feature = "amortize")]
+pub(crate) use indexmap_amortized::{map::Entry, IndexMap as MapImpl};
```
mlodato517, line 8: Seeing how concentrated this change is is crazy cool. Having this `MapImpl` idea seems super useful in this way. General question - did you plan that from the start? Or did that become clear to you the first time you needed to sub in another implementation? If that was something you planned from the start - was it just something clever you thought of or are there some general ideas stored anywhere that are useful for people to consider when building things? This seems like a cool thing to know about when building a lib.

I know this PR isn't a Q&A but I'm not sure what other medium to use to pick your brain 🤔 Can definitely take this offline somewhere if more convenient/keeps the PR discussion clean + focused

jonhoo: Happy to do it here :)
No, I didn't think of this from the start — it just seemed like a handy way to avoid repeating the same config all over the place once I needed more than one impl!

mlodato517: Coolio - good to know.

jonhoo:
This has been folded into https://github.com/jonhoo/evmap and merged in `11.0.0-alpha.6`.

PR #72, opened 2020-11-16
Upgrade to Bytes 0.0.6

benbromhead:
Quick update for evmap to use latest version of bytes

`cargo test` makes this look simple enough

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/rust-evmap/72)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,7 @@ eviction = ["indexed", "rand"]
 indexmap = { version = "1.1.0", optional = true }
 smallvec = "1.0.0"
 hashbag = "0.1.2"
-bytes = { version = "0.5", optional = true }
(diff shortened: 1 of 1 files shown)


jonhoo:
Thank you! Sadly, this is a breaking change, so I'll have to hold off on merging this until the next major release. I _think_ that'll come in not too long though :)

benbromhead:
All good I'd just left the version as `version = "11.0.0-alpha.1"` based on master, but yeah its def a major version change. 

I ended up moving my other deps back to bytes=0.5 anyway (:

jonhoo:
Heads up that this will be fixed by #83, since it no longer uses the `ShallowCopy` trait, and thus we don't need to implement anything for `Bytes` at all!

PR #86, opened 2020-12-15
Implement Extend for WriteHandle

glittershark:
Add a impl for Extend<Op> to WriteHandle, to append multiple operations
from an iterator from the oplog at once. In addition to convenience,
this has the potential to be more efficient by avoiding reallocating the
vector for the oplog multiple times.

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/rust-evmap/86)

### DIFF
--- a/left-right/src/write.rs
+++ b/left-right/src/write.rs
@@ -299,18 +299,7 @@ where
     ///
     /// Its effects will not be exposed to readers until you call [`publish`](Self::publish).
     pub fn append(&mut self, op: O) -> &mut Self {
-        if self.first {
-            // Safety: we know there are no outstanding w_handle readers, since we haven't
-            // refreshed ever before, so we can modify it directly!
-            let mut w_inner = self.raw_write_handle();
-            let w_inner = unsafe { w_inner.as_mut() };
-            let r_handle = self.enter().expect("map has not yet been destroyed");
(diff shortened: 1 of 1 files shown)


jonhoo:
That's a great idea, love it! One little nit inline and then I'm happy to merge this.

left-right/src/write.rs
```diff
@@ -299,18 +299,7 @@ where
...
-            let mut w_inner = self.raw_write_handle();
-            let w_inner = unsafe { w_inner.as_mut() };
-            let r_handle = self.enter().expect("map has not yet been destroyed");
-            // Because we are operating directly on the map, and nothing is aliased, we do want
-            // to perform drops, so we invoke absorb_second.
-            Absorb::absorb_second(w_inner, op, &*r_handle);
-        } else {
-            self.oplog.push_back(op);
-        }
+        self.extend(vec![op]);
```
jonhoo, line 302: I think this should probably use `std::iter::once` so we don't have to allocate a `Vec`.

glittershark: ah, yes, I knew something like that had to exist!

jonhoo (approved):

jonhoo:
Thanks!

jonhoo:
Published in `left-right 0.11` and `evmap 11.0.0-alpha.5` :tada:

Issue #85, opened 2020-12-12
can DropBehavior have an associated constant instead of a function


Eh2406:
```rust
pub trait DropBehavior {
    const DO_DROP: bool;
}
```

That feels less complicated than a method that returns a constant.

jonhoo:
Oh, interesting, I didn't even know that traits could have associated constants now. Absolutely! Want to submit a PR? I don't think it'll make much difference in practice, but it is maybe a little nicer. It is a breaking change to `left-right`, but I think that's fine since it's so new and still pre-`1.0`.

One thing I do worry about with `DropBehavior` is that the compiler gets smart enough that it recognizes that the `Drop` implementation for `Aliased<_, NoDrop>` is a no-op, and _therefore_ decides to lay it out differently in memory. I don't _think_ that's something it does, but it is something that making it a constant will make it more likely optimization. But that said, if that optimization is possible, it would make what we're doing unsound regardless of whether it's an associated const or an `fn`, so the change is likely a good one anyway.

Eh2406:
> Want to submit a PR?

It seems fairly straight forward, so I think I'd prefer to leave it open for someone with less experience to pick up. I am open to mentoring or pair programming or answering questions for anyone that does want to give it a try.

ShaneEverittM:
> > Want to submit a PR?
> 
> It seems fairly straight forward, so I think I'd prefer to leave it open for someone with less experience to pick up. I am open to mentoring or pair programming or answering questions for anyone that does want to give it a try.

I'll give it a shot. I'm confident with Rust, not so much with open source development.

Eh2406:
Let us know how we can help!

ShaneEverittM:
Will do!

ShaneEverittM:
PR made.

ShaneEverittM:
Since evmap depends on the crates.io version of left-right, how should I handle propagating this change into evmap?

Eh2406:
I would think that you would change [this line](https://github.com/jonhoo/rust-evmap/blob/master/evmap/Cargo.toml#L29) to `left-right = { path = "../left-right", version = "0.11.0"  }`. That tells Cargo to use the path for development but depend on the next crates.io version on publish. But I don't know why that is set up to depend on the version from `crates.io` now, so we may want @jonhoo's explanation of the current set up.

ShaneEverittM:
Yeah, that works fine. All tests pass under new implementation, holding off on any more commits though until we get some feedback on the structure.

[...]

PR #87, opened 2020-12-16
Change DropBehavior to have associated const instead of function.

ShaneEverittM:
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/rust-evmap/87)

### DIFF
--- a/evmap/Cargo.toml
+++ b/evmap/Cargo.toml
@@ -26,7 +26,7 @@ indexmap = { version = "1.1.0", optional = true }
 smallvec = "1.0.0"
 hashbag = "0.1.2"
 rand = { version = "0.7", default-features = false, features = ["alloc"], optional = true }
-left-right = "0.10.0"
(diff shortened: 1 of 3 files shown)


Eh2406:
cc #85

jonhoo:
Ah, excellent, thank you! This is going to require a breaking change for `left-right`, so could you please update `evmap` to have a `path` dependency on `left-right` as well, and then fix the resulting error in `evmap` (specifically its `impl DropBehavior` blocks also have to change).

jonhoo (approved):
Looks great, thank you! I'll do a release of `left-right` shortly, since I think we should get this change out sooner rather than later.

ShaneEverittM:
Cool! First PR ever!

jonhoo:
Published in `left-right 0.11` and `evmap 11.0.0-alpha.5` :tada:

PR #89, opened 2021-01-27
Add test case for append on write handle.

grv07:
Hey,
I have added a test case for the append function of the write handle.
This test will check oplog and other parameters "ex: first" behavior, before and after publishing.

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/left-right/89)


jonhoo (approved):
Seems reasonable to me, thanks!

PR #90, opened 2021-01-29
Added test case for the wait function of write handler.

grv07:
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/left-right/90)

### DIFF
--- a/src/write.rs
+++ b/src/write.rs
@@ -515,10 +515,6 @@ mod tests {
         // join to make sure that wait must return after the progress/increment
         // of held_epoch.
         let _ = wait_handle.join();
-
-        // check that writes waiting state is set to false after wait returns.
-        let is_waiting_v = is_waiting.load(Ordering::SeqCst);
-        assert_eq!(false, is_waiting_v);
     }
 
     #[test]
--- a/src/write.rs
+++ b/src/write.rs
@@ -435,6 +435,9 @@ struct CheckWriteHandleSend;
 #[cfg(test)]
 mod tests {
     use crate::CounterAddOp;
+    use slab::Slab;
+    use std::sync::atomic::{AtomicUsize, Ordering};
+    use std::sync::{Arc, Mutex};
 
     #[test]
     fn append_test() {
@@ -448,6 +451,41 @@ mod tests {
         w.append(CounterAddOp(2));
         w.append(CounterAddOp(3));
         assert_eq!(w.oplog.len(), 2);
+    }
+
+    #[test]
+    fn wait_test() {
+        use std::thread;
+
+        let (mut w, _r) = crate::new::<i32, _>();
+
+        // Case 1: If epoch is set to default.
+        let test_epoch: crate::Epochs = Default::default();
+        let mut test_epoch = test_epoch.lock().unwrap();
+        w.wait(&mut test_epoch);
+
+        // Case 2: If one of the reader is still reading(epoch is odd and count is same as in last_epoch)
+        // and wait has been called.
+        let holded_epoch = Arc::new(AtomicUsize::new(1));
+
+        w.last_epochs = vec![2, 2, 1];
+        let mut epoch_slab = Slab::new();
+        epoch_slab.insert(Arc::new(AtomicUsize::new(2)));
+        epoch_slab.insert(Arc::new(AtomicUsize::new(2)));
+        epoch_slab.insert(Arc::clone(&holded_epoch));
+
+        // A new thread act as a reader thread that will increase epoch by 1 and break wait cycle.
+        let move_val = Arc::clone(&holded_epoch);
+        thread::spawn(move || {
+            move_val.fetch_add(1, Ordering::SeqCst);
+        });
+
+        let test_epoch = Arc::new(Mutex::new(epoch_slab));
+        let mut test_epoch = test_epoch.lock().unwrap();
+        w.wait(&mut test_epoch);
+
+        let count = holded_epoch.load(Ordering::Relaxed);
+        assert_eq!(count, 2);
     }
 
     #[test]
--- a/src/write.rs
+++ b/src/write.rs
@@ -144,7 +144,7 @@ where
 
         #[cfg(test)]
         {
-            self.is_waiting.swap(true, Ordering::Relaxed);
+            self.is_waiting.store(true, Ordering::Relaxed);
         }
         // we're over-estimating here, but slab doesn't expose its max index
         self.last_epochs.resize(epochs.capacity(), 0);
@@ -191,7 +191,7 @@ where
         }
         #[cfg(test)]
         {
-            self.is_waiting.swap(false, Ordering::Relaxed);
+            self.is_waiting.store(false, Ordering::Relaxed);
         }
     }
 
@@ -491,7 +491,7 @@ mod tests {
 
         let is_waiting = Arc::clone(&w.is_waiting);
 
-        // check for writer waiting state before calling wait.
+        // check writers waiting state before calling wait.
         let is_waiting_v = is_waiting.load(Ordering::Relaxed);
         assert_eq!(false, is_waiting_v);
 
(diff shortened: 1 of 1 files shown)


grv07:
Hello,
I have added a test case for the wait function of the write handler.

This test checks a case where one of the readers is still reading and the hander calls the wait fn.
I try to simulate the read handle behaviour by creating a new thread that will increase the epoch by one when scheduled.
Thanks.

jonhoo:
Hmm, I don't think this test does quite what you want it to. In particular, the thread that increments may end up running before `wait` is even called, in which case `wait` won't wait at all. There isn't _really_ a good way to fix that problem except by creating a `loom` test to test both interleavings, which I know is something @mlodato517 was looking at a little while back.

grv07:
Yes, I understand your point.
Apart from loom do you think park_timeout will be good fit for this problem?
If yes should I update it with park_timeout ?

jonhoo:
`park_timeout` could work too! You'd probably want to first `park_timeout` _without_ the reader allowing the writer to make progress, check that the writer did in fact not make progress, then `park`, then progress the reader, and then check that the writer was then unblocked.

src/write.rs
```diff
@@ -450,6 +465,48 @@ mod tests {
...
+        w.last_epochs = vec![2, 2, 1];
+        let mut epochs_slab = Slab::new();
+        epochs_slab.insert(Arc::new(AtomicUsize::new(2)));
+        epochs_slab.insert(Arc::new(AtomicUsize::new(2)));
+        epochs_slab.insert(Arc::clone(&holded_epoch));
+
+        // A new thread act as a reader thread that will increase epoch by 1 and break wait cycle.
+        let move_holded_epoch = Arc::clone(&holded_epoch);
+        let progress_handle = thread::spawn(move || {
+            thread::park_timeout(Duration::from_secs(2));
```
jonhoo, line 492: Ah, if this is what you wanted to do, you'll probably want to use `sync::Barrier` instead so that you don't end up with a race condition in the test :)

src/write.rs
```diff
@@ -179,6 +187,10 @@ where
                     continue 'retry;
                 }
             }
+            #[cfg(test)]
```
jonhoo, line 190: Maybe stick this below the loop rather than just before the `break`?

src/write.rs
```diff
@@ -450,6 +465,48 @@ mod tests {
...
+    #[test]
+    fn wait_test() {
+        use std::thread;
+        use std::time::Duration;
+        let (mut w, _r) = crate::new::<i32, _>();
+
+        // Case 1: If epoch is set to default.
+        let test_epochs: crate::Epochs = Default::default();
+        let mut test_epochs = test_epochs.lock().unwrap();
+        w.wait(&mut test_epochs);
```
jonhoo, line 477: It'd be good to document here why you expect `wait` to return immediately.

src/write.rs
```diff
@@ -450,6 +465,48 @@ mod tests {
...
+
+        // Case 2: If one of the reader is still reading(epoch is odd and count is same as in last_epoch)
+        // and wait has been called.
+        let holded_epoch = Arc::new(AtomicUsize::new(1));
```
jonhoo, line 481:
```suggestion
        let held_epoch = Arc::new(AtomicUsize::new(1));
```

src/write.rs
jonhoo: I think this may also end up nicer if you make the spawned thread do the call to `wait` instead. Then the main test code can first check that it's _not_ waiting before the barrier, then check that it _is_ waiting shortly after, then check that it stays that way after a call to `yield_now`, then bump the `held_epoch`, and then check that it can join the spawned thread.

src/write.rs
```diff
@@ -450,6 +465,66 @@ mod tests {
...
+
+        w.last_epochs = vec![2, 2, 1];
+        let mut epochs_slab = Slab::new();
+        epochs_slab.insert(Arc::new(AtomicUsize::new(2)));
+        epochs_slab.insert(Arc::new(AtomicUsize::new(2)));
+        epochs_slab.insert(Arc::clone(&held_epoch));
+
+        let barrier = Arc::new(Barrier::new(2));
+
+        let is_waiting = Arc::clone(&w.is_waiting);
```
jonhoo, line 492: You can probably just assert that this is `false` here rather than doing it in both threads before the barrier wait below. Arguably you don't even need that pre-assert.

src/write.rs
```diff
@@ -450,6 +465,66 @@ mod tests {
...
+        let c = Arc::clone(&barrier);
```
jonhoo, line 493: I'd call this `barrier2` or something like it so it's clearer below that the thread is also waiting on the barrier. Single-character variable names tend to make code hard to follow.

src/write.rs
```diff
@@ -450,6 +465,66 @@ mod tests {
...
+
+        let m_is_waiting = Arc::clone(&w.is_waiting);
+        let test_epochs = Arc::new(Mutex::new(epochs_slab));
+
+        let wait_handle = thread::spawn(move || {
+            let is_waiting = m_is_waiting.load(Ordering::Relaxed);
+            assert_eq!(false, is_waiting);
+            let mut test_epochs = test_epochs.lock().unwrap();
```
jonhoo, line 501: Let's take this lock after the barrier wait instead of before. Feels wrong to hold a lock while waiting on a barrier.

src/write.rs
```diff
@@ -450,6 +465,66 @@ mod tests {
...
+        let is_waiting_v = is_waiting.load(Ordering::Relaxed);
+        assert_eq!(false, is_waiting_v);
+
+        barrier.wait();
+        // make sure that writer wait() will call first then change the held epoch.
+        thread::yield_now();
+
+        let is_waiting_v = is_waiting.load(Ordering::Relaxed);
+        // check if is_waiting is set to true after calling wait.
+        assert_eq!(true, is_waiting_v);
```
jonhoo, line 515: I don't think you're guaranteed that this is `true` even after a call to `yield_now`. You'll probably instead want `yield_now` in a loop until `is_waiting` is `true`. And then yield once more and check that it's still true.

src/write.rs
```diff
@@ -450,6 +465,66 @@ mod tests {
...
+
+        held_epoch.fetch_add(1, Ordering::SeqCst);
+
+        let _ = wait_handle.join();
```
jonhoo, line 519: A comment here saying why we now expect the wait to have made progress would be good.

src/write.rs
```diff
@@ -128,16 +130,22 @@ where
...
     }
 
-    fn wait(&mut self, epochs: &mut MutexGuard<'_, slab::Slab<Arc<atomic::AtomicUsize>>>) {
+    fn wait(&mut self, epochs: &mut MutexGuard<'_, slab::Slab<Arc<AtomicUsize>>>) {
         let mut iter = 0;
         let mut starti = 0;
 
+        #[cfg(test)]
+        {
+            self.is_waiting.swap(true, Ordering::Relaxed);
```
jonhoo, line 147: I believe this can just be a `store`?

src/write.rs
```diff
@@ -181,6 +189,10 @@ where
             }
             break;
         }
+        #[cfg(test)]
+        {
+            self.is_waiting.swap(false, Ordering::Relaxed);
```
jonhoo, line 194: Same here — I think this can be a `store`

src/write.rs
```diff
@@ -450,6 +465,73 @@ mod tests {
...
+        let wait_handle = thread::spawn(move || {
+            barrier2.wait();
+            let mut test_epochs = test_epochs.lock().unwrap();
+            w.wait(&mut test_epochs);
+        });
+
+        barrier.wait();
+
+        // make sure that writer wait() will call first only then update the held epoch.
+        'retry: loop {
```
jonhoo, line 509: I wonder, can this just be a `while` loop?
```rust
while !is_waiting.load(Ordering::Relaxed) {
    thread::yield_now();
}
```

src/write.rs
```diff
@@ -450,6 +465,73 @@ mod tests {
...
+            let is_waiting_v = is_waiting.load(Ordering::Relaxed);
+            if is_waiting_v {
+                thread::yield_now();
+                break 'retry;
+            }
+        }
+
+        let is_waiting_v = is_waiting.load(Ordering::Relaxed);
+        // check if is_waiting is set to true after calling wait.
+        assert_eq!(true, is_waiting_v);
```
jonhoo, line 520: I think you can remove this — you wouldn't get to this `assert` unless `is_waiting_v` is already true.

src/write.rs
```diff
@@ -450,6 +465,73 @@ mod tests {
...
+
+        held_epoch.fetch_add(1, Ordering::SeqCst);
+
+        // join to make sure that wait fn must return after the progress/increment
+        // of held_epoch.
+        let _ = wait_handle.join();
+
+        let count = held_epoch.load(Ordering::Relaxed);
+        assert_eq!(count, 2);
```
jonhoo, line 529: What is this assertion checking?

src/write.rs
```diff
@@ -505,29 +505,19 @@ mod tests {
...
-            if is_waiting_v {
-                thread::yield_now();
-                break 'retry;
-            }
         }
-
-        let is_waiting_v = is_waiting.load(Ordering::Relaxed);
-        // check if is_waiting is set to true after calling wait.
-        assert_eq!(true, is_waiting_v);
+        thread::yield_now();
```
jonhoo, line 512: Why is this extra yield needed?

jonhoo:
I think we're almost there now!

src/write.rs
```diff
@@ -450,6 +465,62 @@ mod tests {
...
+            thread::yield_now();
+        }
+
+        held_epoch.fetch_add(1, Ordering::SeqCst);
+
+        // join to make sure that wait must return after the progress/increment
+        // of held_epoch.
+        let _ = wait_handle.join();
+
+        // check that writes waiting state is set to false after wait returns.
```
jonhoo, line 519: I think I'd just remove this assertion — it doesn't really add much, and anything that's in the test that's not relevant _to_ the test might confuse later readers :)

jonhoo:
Oh, and if you can merge from `master`, we should get CI testing and coverage analysis for this PR too!

jonhoo (approved):

jonhoo:
Awesome, thank you for sticking with it!

PR #91, opened 2021-02-05
Add infrastructure for `loom` tests

mlodato517:
**This Commit**
Sets up testing with the `loom` infrastructure by doing three things:
1. creating a `mod sync` that exports `std` or `loom` versions of
concurrency primitives depending on the configuration flags (as
recommended by [the `loom` docs][1]).
2. changing all imports of the various std::sync::* concurrency
primitives to crate::sync::* so that we can toggle to `loom`
primitives from `std` lib primitives when testing
3. writing a bare-bones loom test for others to build upon :-)

Related to #17 

[1]: https://github.com/tokio-rs/loom#structuring-tests

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/left-right/91)

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -52,6 +52,17 @@ jobs:
      - script: |
            env RUSTFLAGS="-Z sanitizer=leak" cargo test --target x86_64-unknown-linux-gnu
        displayName: cargo -Z sanitizer=leak test
+ - job: loom
+   displayName: "Loom tests"
+   pool:
+     vmImage: ubuntu-latest
+   steps:
+     - template: install-rust.yml@templates
+       parameters:
+         rust: nightly
+     - script: |
+           env RUSTFLAGS="--cfg loom" cargo test --test loom
+       displayName: cargo test loom
 
 resources:
   repositories:
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -61,7 +61,7 @@ jobs:
        parameters:
          rust: nightly
      - script: |
-           env RUSTFLAGS="--cfg loom" cargo test --test loom
+           env LOOM_MAX_PREEMPTIONS=3 RUSTFLAGS="--cfg loom" cargo test --test loom
        displayName: cargo test loom
 
 resources:
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -174,9 +174,11 @@
 )]
 #![allow(clippy::type_complexity)]
 
-use std::sync::{atomic, Arc, Mutex};
+mod sync;
 
-type Epochs = Arc<Mutex<slab::Slab<Arc<atomic::AtomicUsize>>>>;
+use crate::sync::{Arc, AtomicUsize, Mutex};
+
+type Epochs = Arc<Mutex<slab::Slab<Arc<AtomicUsize>>>>;
 
 mod write;
 pub use crate::write::WriteHandle;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -296,6 +296,3 @@ where
     let w = WriteHandle::new(T::default(), epochs, r.clone());
     (w, r)
 }
-
-#[cfg(test)]
-mod utilities;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -298,21 +298,4 @@ where
 }
 
 #[cfg(test)]
-struct CounterAddOp(i32);
-
-#[cfg(test)]
-impl Absorb<CounterAddOp> for i32 {
-    fn absorb_first(&mut self, operation: &mut CounterAddOp, _: &Self) {
-        *self += operation.0;
-    }
-
-    fn absorb_second(&mut self, operation: CounterAddOp, _: &Self) {
-        *self += operation.0;
-    }
-
-    fn drop_first(self: Box<Self>) {}
-
-    fn sync_with(&mut self, first: &Self) {
-        *self = *first
-    }
-}
+mod utilities;
--- a/src/read.rs
+++ b/src/read.rs
@@ -211,7 +211,7 @@ impl<T> ReadHandle<T> {
 ///
 /// // the line below will not compile as ReadHandle does not implement Sync
 ///
-/// is_<ReadHandle<u64>>()
+/// is_sync::<ReadHandle<u64>>()
 /// ```
 ///
 /// But, it can be sent across threads:
--- a/src/read.rs
+++ b/src/read.rs
@@ -1,10 +1,8 @@
+use crate::sync::{fence, Arc, AtomicPtr, AtomicUsize, Ordering};
 use std::cell::Cell;
 use std::fmt;
 use std::marker::PhantomData;
 use std::ptr::NonNull;
-use std::sync::atomic;
-use std::sync::atomic::AtomicPtr;
-use std::sync::{self, Arc};
 
 // To make [`WriteHandle`] and friends work.
 #[cfg(doc)]
@@ -39,9 +37,9 @@ pub use factory::ReadHandleFactory;
 /// a [`ReadHandleFactory`]. Note, however, that creating a new handle through either of these
 /// mechanisms _does_ take a lock, and may therefore become a bottleneck if you do it frequently.
 pub struct ReadHandle<T> {
-    pub(crate) inner: sync::Arc<AtomicPtr<T>>,
+    pub(crate) inner: Arc<AtomicPtr<T>>,
     pub(crate) epochs: crate::Epochs,
-    epoch: sync::Arc<sync::atomic::AtomicUsize>,
+    epoch: Arc<AtomicUsize>,
     epoch_i: usize,
     enters: Cell<usize>,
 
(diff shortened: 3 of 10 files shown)


src/lib.rs
```diff
@@ -174,9 +174,11 @@
 )]
 #![allow(clippy::type_complexity)]
 
-use std::sync::{atomic, Arc, Mutex};
+mod sync;
 
-type Epochs = Arc<Mutex<slab::Slab<Arc<atomic::AtomicUsize>>>>;
```
mlodato517, line 179: Question - how do we feel about changing `atomic::X` imports to just be `X`? Do we want to maintain the module for readability/discoverability?

src/read.rs
```diff
@@ -216,7 +211,7 @@ impl<T> ReadHandle<T> {
 ///
 /// // the line below will not compile as ReadHandle does not implement Sync
 ///
-/// is_sync::<ReadHandle<u64>>()
+/// is_<ReadHandle<u64>>()
```
mlodato517, line 214: Whoops! Need to fix this :-)

mlodato517 (commented):
Note too that I didn't touch `src/aliasing` because of the `?Sized` trait bound in the implementations for the various smart pointers. Not sure how badly we want that.

mlodato517:
Whoops I probably need to add loom tests to the CI

tests/loom.rs
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    #[test]
+    fn read_before_publish() {
+        loom::model(|| {
+            let (mut w, r) = left_right::new::<i32, _>();
+            w.append(CounterAddOp(1));
+            w.publish();
+
+            let val = *r.enter().unwrap();
```
mlodato517, line 36: I think this read should be after the thread spawn but before the join and I should remove the publish from the thread closure?

src/read.rs
mlodato517: cd21a0f43ce2de29ce3330aab22a7bba87b05c48

tests/loom.rs
mlodato517: Huh I updated this with 753fd8704bc7959b73249a35abe6c00524c57983 but now, when running loom tests I get:
```
thread 'loom_tests::read_before_publish' panicked at 'Model exeeded maximum number of branches. 
```
which I think is something I ran into [last time](https://github.com/jonhoo/left-right/pull/62#discussion_r468213451) ... not sure what's going on here 🤔

src/lib.rs
jonhoo: That seems fine to me!

tests/loom.rs
```diff
@@ -0,0 +1,48 @@
+#[cfg(loom)]
+#[cfg(test)]
+mod loom_tests {
+    use left_right::{Absorb, ReadGuard};
+    use loom::sync::atomic::AtomicUsize;
+    use loom::sync::atomic::Ordering::{Acquire, Relaxed, Release};
+    use loom::sync::Arc;
+    use loom::thread;
+
+    // TODO Would love to not re-define this!
```
jonhoo, line 10: I think you could do that by making a subdirectory `common`, stick the struct + impls in `common/mod.rs`, and then place `mod common` at the top of this file.

jonhoo:
I think here
https://github.com/jonhoo/left-right/blob/753fd8704bc7959b73249a35abe6c00524c57983/src/write.rs#L169-L173

You'll want to just immediately `yield_now` under `loom`.

tests/loom.rs
jonhoo: The updated code looks right to me. What value are you running with for `MAX_PREEMPTIONS`?

jonhoo:
This is awesome, thanks for digging into it!

tests/loom.rs
mlodato517: Hmmm, sorry, I'm confused there 😅. Since this test is in `tests/loom.rs` and not under `src`, where would `common` go? Under `src` or `tests` or neither and be at the same level? If I'm going to have `mod common;` in this file then I imagine it has to be under `tests/` as well but then can I access it in `src/` tests? Or should I move the `loom` tests to be under `src` somewhere?

jonhoo: Ah, yes, if you also want to use it in unit tests in `src/`, you have to play a bit more of a trick. Specifically, you'd have to use the [`#[path]` attribute](https://doc.rust-lang.org/nightly/reference/items/modules.html#the-path-attribute) on the `mod` statement in `src/` to bring in the type from `tests/` (or vice-versa) or the [`include!` macro](https://doc.rust-lang.org/std/macro.include.html) to just directly dump in the source. It gets further complicated by the fact that the path to the trait is different in the two places. I think the way to go is probably to use `include!` and then place the `use` for the trait just above the call to `include!`.

mlodato517: Okay, great, I've attempted this with 606d6c3046379a9d0b6b311871682499b307113f. Lemme know if that's what you had in mind! I tried to follow https://stackoverflow.com/a/44545091/2891835 but I think there was a circular dependency since `left_right` had the utilities as a test dependency but the utilities had `left_right` as a dependency so it could implement `Absorb` 🤷

jonhoo: Ah, almost. I was proposing you use `include!` in _both_ `src/` and `test/`

mlodato517: Ah! So like 98b54b76a0694e2e910bd33857e007148a207095?

tests/loom.rs
mlodato517: Hmmm so I locally added the `loom::thread::yield_now` (good call!) but I'm realizing that this test may not be deterministic - if the write handle is dropped before the read happens, the reader will return `None`. Is that right?

mlodato517: So I think the more safe test is to read the value in the thread and do the publish between the spawn and join. Then, we can ensure the write handle stays alive long enough to validate the test by using it after the join/assert/something?

mlodato517: Made those adjustments in b7442f68147bd7bb505f56615f6e3f3e2489fc6c. Lemme know if that's right! If so, I think the last step is just to update CI

src/write.rs
```diff
@@ -176,6 +172,9 @@ where
                         thread::yield_now();
                     }
 
+                    #[cfg(loom)]
+                    loom::thread::yield_now();
```
jonhoo, line 176: I think we should also make the `if` just above be `cfg(not(loom))`

tests/loom.rs
```diff
@@ -0,0 +1,38 @@
...
+    use loom::sync::atomic::{AtomicI32, Ordering};
+    use loom::sync::Arc;
+    use loom::thread;
+
+    #[test]
+    fn read_before_publish() {
+        loom::model(|| {
+            let (mut w, r) = left_right::new::<i32, _>();
+            let answer = Arc::new(AtomicI32::new(0));
+            let answer_clone = answer.clone();
```
jonhoo, line 18: You don't need this — instead, you can just return `val` from inside the spawned thread, and then get the return value out of `jh.join().unwrap()`.

tests/loom.rs
```diff
@@ -0,0 +1,38 @@
...
+            let jh = thread::spawn(move || {
+                let val = *r.enter().unwrap();
+                answer_clone.store(val, Ordering::SeqCst);
+            });
+            w.publish();
+            w.append(CounterAddOp(1));
+
+            jh.join().unwrap();
+
+            // This is just to ensure the write handle is not dropped before the reader reads
```
jonhoo, line 32: I'm not sure I understand this comment? `w` won't be dropped until the end of the scope anyway..

tests/loom.rs
mlodato517: TIL! That's awesome. Updated this in a50870be1a4e7e5577084a8e109790c0373cd02e

tests/loom.rs
mlodato517: That's a very good point 😅. I think I was afraid rust would think like, "Well this isn't used anymore, might as well clean up early" or something crazy. Updated in cc8711d6623a652b9b3327edf9c7a1269df77d29

src/write.rs
mlodato517: Makes sense to me! Updated in 4342b2262507916c2f437a8693137a6d0bed0635. Not sure if this would like better with like an `if cfg!(loom) { } else { }` - lemme know your preference there (or maybe that's not possible?). Also, this now causes some unused variable complaints when running with the loom configuration. I assume that's fine but I can chase those down if you want!

jonhoo: I think I'd just surround the original `if else` in an `if cfg!(loom)`, and then not place the `#[cfg(loom)]` in an `else` block but leave it like you have here.

mlodato517: Hmmm, sorry, confused again :sweat_smile: do you mean like:
```rust
if !cfg!(loom) {
  if iter != 20 { ... } else { ... }
}

#[cfg(loom)]
loom::thread::yield_now();
```

jonhoo: Yes indeed. That should get rid of the "unused" warnings :)

mlodato517: Ah, clever! Added in 62a4de0221322e5bdf6f6ad71df05d8f113bba22

azure-pipelines.yml
```diff
@@ -51,6 +51,17 @@ jobs:
...
+    - job: loom
+      displayName: "Loom tests"
+      pool:
+        vmImage: ubuntu-latest
+      steps:
+        - template: install-rust.yml@templates
+          parameters:
+            rust: nightly
+        - script: |
+              env RUSTFLAGS="--cfg loom" cargo test --test loom
```
jonhoo, line 63: I wonder if this'll have to set `LOOM_MAX_PREEMPTIONS` as well or the tests will run _forever_. See https://github.com/tokio-rs/loom#handling-large-models

mlodato517: Ah, interesting. I assume this is future proofing the CI job? Considering that it currently runs on my laptop "instantaneously"

mlodato517: Either way, added in 04a8c15fdf51e136815dbb146ea7f1b95d51cd41!

jonhoo:
This is looking good now! I just pushed a commit that _should_ fix CI, so if you merge that in we'll see if it flies!

jonhoo (approved):

mlodato517:
Rebased 👍

jonhoo:
It'd be _really_ nice to get tarpaulin coverage for the loom tests too. That should help a lot! I _think_ that should be doable (@xd009642 we basically want to _also_ run a particular test suite under `tarpaulin` with a particular `--cfg` passed in), but we can save that for a later PR.

jonhoo:
@mlodato517 Support for `RUSTFLAGS` just landed in tarpaulin! If you want to take a stab at running tarpaulin coverage of loom tests, I'd probably start by inlining [`coverage.yml`](https://github.com/crate-ci/azure-pipelines/blob/master/coverage.yml) into our `azure-pipelines.yml`, and then add+tweak a [tarpaulin config file](https://github.com/xd009642/tarpaulin#config-file) to run both the standard tests and loom. If you run into issues, I'm sure @xd009642 would love to hear about them!

xd009642:
FYI it's only in tarpaulin when you install it from the develop branch, it's not been released onto crates.io in anyway. Or the docker images with develop in the tags (if they've built and pushed yet)

PR #92, opened 2021-04-05
Fix leftover brackets in docs

nicolaiunrein:
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/left-right/92)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -48,7 +48,7 @@
 //! points to a `T` that the readers access in order to read data. Every time a read has accessed
 //! the pointer, they increment a local epoch counter, and they update it again when they have
 //! finished the read. When a write occurs, the writer updates the other `T` (for which there are
-//! no readers), and also stores a copy of the change in a log. When [`WriteHandle::publish`) is
+//! no readers), and also stores a copy of the change in a log. When `WriteHandle::publish` is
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -48,7 +48,7 @@
 //! points to a `T` that the readers access in order to read data. Every time a read has accessed
 //! the pointer, they increment a local epoch counter, and they update it again when they have
 //! finished the read. When a write occurs, the writer updates the other `T` (for which there are
-//! no readers), and also stores a copy of the change in a log. When [`WriteHandle::publish`) is
+//! no readers), and also stores a copy of the change in a log. When `WriteHandle::publish` is
```
jonhoo, line 51: Ah, these weren't actually leftover brackets, but a (failed) attempt at [intra-doc links](https://doc.rust-lang.org/rustdoc/linking-to-items-by-name.html). I _think_ the right fix here is to replace the `)` with a `]`, and then `cargo doc` should run without complaining!

nicolaiunrein: Oh I thought it used to be a `[]()` link. Didn’t even think of  the `[]` syntax.

jonhoo (approved):
Thank you! :tada:

PR #94, opened 2021-05-04
add WriteHandle::take

pYtoner:
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/left-right/94)


pYtoner:
There might be a better way to signal to the drop that it needs to not do its thing. But a bool might be worth the clearness.

pYtoner:
I am also not sure how the documentation should mention that the drop second should be called if it is needed or if it should at all.

PR #96, opened 2021-05-04
add WriteHandle::take

pYtoner:
There might be a less clear way to let the Drop know that it needs to do things differently but a bool is not much overhead.

I am not sure how the usage of drop second should be documented. But what I wrote should be a good start I think.

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/left-right/96)

### DIFF
--- a/src/write.rs
+++ b/src/write.rs
@@ -38,6 +38,8 @@ where
     first: bool,
     /// A publish has happened, but the two copies have not been synchronized yet.
     second: bool,
+    /// If we call `Self::take` the drop needs to be different.
+    taken: bool,
 }
 
 // safety: if a `WriteHandle` is sent across a thread boundary, we need to be able to take
@@ -75,6 +77,12 @@ where
     T: Absorb<O>,
 {
     fn drop(&mut self) {
+        // If we use `Self::take` then we cannot drop both r_handle and w_handle so
+        // a different cleanup happens there.
+        if self.taken {
+            return;
+        }
+
         use std::ptr;
 
         // first, ensure both copies are up to date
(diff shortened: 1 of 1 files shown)


src/write.rs
```diff
@@ -75,6 +77,12 @@ where
     T: Absorb<O>,
 {
     fn drop(&mut self) {
+        // If we use `Self::take` then we cannot drop both r_handle and w_handle so
+        // a different cleanup happens there.
+        if self.taken {
+            return;
```
pYtoner, line 83: There is a test coverage warning that says this wasn't covered by tests. But it gets called when the take functions returns.

pYtoner:
Should the version of the crate be changed also?

jonhoo (changes requested):
Ah, no, no need to update the version number -- I'll take care of that.

I'd like to see this take a slightly different approach that lets us share the code between `drop` and `take`. Specifically, what I'm thinking is to add a `take_inner(&mut self) -> Option<T>` that does all the stuff `take`/`drop` currently does. Then, call that from both `drop` and `take`, with `drop` just dropping the return value. Does that make sense? This code is hairy enough that I don't want it repeated if we can avoid it.

pYtoner:
That seems like a good approach. I'll do it today.

jonhoo:
Oh, by the way, you don't need to open a new PR for each change. You can just push another commit to the branch you made the commit from :)

pYtoner:
Oh that makes it much easier. Thanks for letting me know. My first time contributing so things are new and fun :)

PR #97, opened 2021-05-05
add WriteHandle::take using take_inner

pYtoner:
As mentioned in https://github.com/jonhoo/left-right/pull/96#pullrequestreview-651799305:
> add a `take_inner(&mut self) -> Option<T>` that does all the stuff `take`/`drop` currently does. Then, call that from both `drop` and `take`, with `drop` just dropping the return value.

But the `take_inner` in this PR returns an `Option<Box<T>>` which I think is better because:
* `T` is already boxed
* drop_second needs a boxed `T`
* `T` might be large so it might be sometimes faster to pass the boxed T

Still not quite sure about my documentation for `take`..

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/left-right/97)

### DIFF
--- a/src/write.rs
+++ b/src/write.rs
@@ -38,6 +38,8 @@ where
     first: bool,
     /// A publish has happened, but the two copies have not been synchronized yet.
     second: bool,
+    /// If we call `Self::take` the drop needs to be different.
+    taken: bool,
 }
 
 // safety: if a `WriteHandle` is sent across a thread boundary, we need to be able to take
@@ -70,16 +72,28 @@ where
     }
 }
 
-impl<T, O> Drop for WriteHandle<T, O>
+impl<T, O> WriteHandle<T, O>
 where
     T: Absorb<O>,
 {
-    fn drop(&mut self) {
+    /// Takes out the inner backing data structure if it hasn't been taken yet. Otherwise returns `None`.
+    ///
+    /// Makes sure that all the pending operations are applied and waits till all the read handles
+    /// have departed. Then it uses `Absorb::drop_first` to drop one of the copies of the data and
+    /// returns the other copy as a boxed value.
+    fn take_inner(&mut self) -> Option<Box<T>> {
         use std::ptr;
+        // Can only take inner once.
+        if self.taken {
+            return None;
+        }
+
+        // Disallow taking again.
+        self.taken = true;
 
         // first, ensure both copies are up to date
-        // (otherwise safely dropping deduplicated data is a pain)
-        if !self.oplog.is_empty() {
+        // (otherwise safely dropping the possibly duplicated w_handle data is a pain)
+        if self.first || !self.oplog.is_empty() {
             self.publish();
         }
         if !self.oplog.is_empty() {
@@ -105,12 +119,24 @@ where
         // safety: w_handle was initially crated from a `Box`, and is no longer aliased.
         Absorb::drop_first(unsafe { Box::from_raw(self.w_handle.as_ptr()) });
 
-        // next we drop the r_handle.
+        // next we take the r_handle and return it as a boxed value.
+        //
         // this is safe, since we know that no readers are using this pointer
         // anymore (due to the .wait() following swapping the pointer with NULL).
         //
         // safety: r_handle was initially crated from a `Box`, and is no longer aliased.
-        Absorb::drop_second(unsafe { Box::from_raw(r_handle) });
+        unsafe { Some(Box::from_raw(r_handle)) }
+    }
+}
+
+impl<T, O> Drop for WriteHandle<T, O>
+where
+    T: Absorb<O>,
+{
+    fn drop(&mut self) {
+        if let Some(inner) = self.take_inner() {
+            Absorb::drop_second(inner);
+        }
     }
 }
 
@@ -133,6 +159,7 @@ where
             refreshes: 0,
             first: true,
             second: true,
+            taken: false,
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/write.rs
```diff
@@ -468,6 +541,23 @@ mod tests {
...
+    fn take_test() {
+        let (mut w, _r) = crate::new_from_empty::<i32, _>(2);
+        w.append(CounterAddOp(1));
+        w.publish();
+        w.append(CounterAddOp(1));
+        assert_eq!(w.take(), 4);
+
+        let (mut w, _r) = crate::new_from_empty::<i32, _>(2);
+        w.append(CounterAddOp(1));
+        // w.publish();
```
jonhoo, line 554: I'd like to see a test with an empty oplog too, just as a sanity check. That is, one with `.append` followed by `.publish`, followed by `.take`.

src/write.rs
```diff
@@ -330,6 +357,52 @@ where
...
         self.w_handle
     }
+
+    /// Returns the backing data structure.
+    ///
+    /// Makes sure that all the pending operations are applied and waits till all the read handles
+    /// have departed. Then it uses `Absorb::drop_first` to drop one of the copies of the data and
+    /// returns the other copy as an owned value.
+    ///
+    /// ## With custom `drop_second`:
```
jonhoo, line 367: I'm very glad you wrote this section — was about to ask about it. I wonder if we want to provide a wrapper type ourselves? Something like:
```rust
struct Absorbing<T: Absorb>(Box<T>);
// impl Deref + DerefMut for Absorbing
impl<T: Absorb> Absorbing<T> {
    /// This is unsafe because you must call Absorb::drop_second
    pub unsafe fn into_box(self) -> Box<T> {
    }
}
impl<T: Absorb> Drop for Absorbing<T> {
    fn drop(&mut self) {
        T::drop_second(self);
    }
}
```
What do you think? The name is maybe a little weird — happy to hear alternatives. I think providing this type makes it much clearer to the caller exactly what they need to do, and tries to enforce that through the type system.

src/write.rs
```diff
@@ -330,6 +357,52 @@ where
...
+    /// #        self.0 += operation.0;
+    /// #    }
+    /// #
+    /// #    fn sync_with(&mut self, first: &Self) {
+    /// #        self.0 = first.0
+    /// #    }
+    /// # }
+    ///
+    /// let original_data = BackingData(10);
+    /// let (write, read) = left_right::new_from_empty::<_, CounterAddOp>(original_data.clone());
```
jonhoo, line 389: `CounterAddOp` shouldn't appear here since it's hidden from the docs.

src/write.rs
```diff
@@ -330,6 +357,52 @@ where
...
+    ///
+    /// let taken_data = write.take();
+    /// assert_eq!(original_data, taken_data);
+    ///
+    /// // make sure it is dropped using `drop_second` if your type requires it.
+    /// Absorb::drop_second(Box::new(taken_data));
```
jonhoo, line 395: I think this should be `BackingData::drop_second(` just for clarity.

src/write.rs
```diff
@@ -70,16 +72,28 @@ where
...
+        }
+
+        // Disallow taking again.
+        self.taken = true;
 
         // first, ensure both copies are up to date
-        // (otherwise safely dropping deduplicated data is a pain)
-        if !self.oplog.is_empty() {
+        // (otherwise safely dropping the possibly duplicated w_handle data is a pain)
+        if self.first || !self.oplog.is_empty() {
```
jonhoo, line 96: Good catch!

jonhoo (changes requested):
This looks much better, thanks for taking the time!

I left a few more comments inline.

src/write.rs
pYtoner: I thought `Absorb::drop_second` was clearer but I'll change it.

src/write.rs
pYtoner: Of course 😃

src/write.rs
pYtoner: It seems like a much safer version. And this also makes the drop implementation a little nicer.

For the naming:
* it only adds extra functionality to `drop_second`
* this is taken owned data

I think good options are: `AbsorbDrop` and `Taken`
I am going to implement it with `AbsorbDrop` for now and we can change it if something better is found.

src/write.rs
pYtoner: I removed that whole part about the safety because it is handled by the smart pointer.

jonhoo (changes requested):
Ah, yes, I like this version much better! Left some mostly minor comments inline.

src/write.rs
```diff
@@ -70,16 +74,86 @@ where
     }
 }
 
-impl<T, O> Drop for WriteHandle<T, O>
+/// A **smart pointer** to an owned backing data structure. This makes sure that the
+/// data is dropped correctly. (Using `Absorb::drop_second`)
+///
+/// Additionally it allows for unsafely getting the inner data out using `.into_box()`
+pub struct AbsorbDrop<T: Absorb<O>, O> {
```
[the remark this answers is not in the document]
pYtoner: `Taken` is better I feel. `Remainder` to me means what is "left over" after left-right is done. It makes me think that the `Remainder` isn't very useful. While `Taken` on the other hand reminds me of `Option::take`. It is up to you but out of the three I prefer `Taken` now.

src/write.rs
```diff
@@ -468,6 +573,31 @@ mod tests {
...
     }
 
+    #[test]
+    fn take_test() {
+        // normal publish then pending operations published by take
+        let (mut w, _r) = crate::new_from_empty::<i32, _>(2);
+        w.append(CounterAddOp(1));
+        w.publish();
+        w.append(CounterAddOp(1));
+        assert_eq!(*w.take(), 4);
```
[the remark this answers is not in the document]
pYtoner: Yes good idea

src/write.rs
```diff
@@ -70,16 +74,89 @@ where
...
+}
+
+impl<T: Absorb<O> + std::fmt::Debug, O> std::fmt::Debug for Taken<T, O> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("Taken")
+            .field(
+                "inner",
+                self.inner
+                    .as_ref()
+                    .expect("inner is only taken in `into_box` which drops self"),
```
jonhoo, line 93: I'd probably refer to this as "consumes self" rather than "drops self", but it's also fine as-is!

jonhoo (approved):
Excellent, thank you for sticking with it!

PR #98, opened 2021-05-30
Adding missing ReadHandleFactory methods

iartem:
It seems `ReadHandleFactory` couldn't be used previously as there were no respective crate functions exported. Not sure if that's intended though.

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/left-right/98)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -297,3 +297,34 @@ where
     let w = WriteHandle::new(T::default(), epochs, r.clone());
     (w, r)
 }
+
+/// Construct a new write randle and read handle factory pair from an empty data structure.
+///
+/// The type must implement `Clone` so we can construct the second copy from the first.
+pub fn new_from_empty_with_factory<T, O>(t: T) -> (WriteHandle<T, O>, ReadHandleFactory<T>)
+where
+    T: Absorb<O> + Clone,
+{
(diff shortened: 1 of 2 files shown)


jonhoo:
Ah, no, the `Factory` isn't a way to construct a _new_ map, it's a way to get a version of a `ReadHandle` that you can share across thread boundaries. The idea is that you construct a map somewhere, create a `ReadHandleFactory` from the `ReadHandle`, and then stick that `ReadHandleFactory` in, say, an `Arc`. Then, in any thread that needs a `ReadHandle`, you call `ReadHandle::handle`.

iartem:
Alright, and how do I create a `ReadHandleFactory` from the `ReadHandle`? Correct me if I'm wrong, but everything is `pub(crate)`.

jonhoo:
You're completely right, seems like a method is indeed missing and I was overly hasty :sweat_smile:. Specifically:
```rust
impl ReadHandle {
    pub fn factory(&self) -> ReadHandleFactory {
        ReadHandleFactor {
            inner: Arc::clone(&self.inner),
            epochs: Arc::clone(&self.epochs),
        }
    }
}
```
Want to submit a PR that adds that method?

iartem:
Done 😄

PR #100, opened 2021-09-14
Allow for compression of the oplog in case of mergeable operations.

Wulf0x67E7:
## What
Adds one associated const and fn each to `Absorb`, as well the new Type `TryCompressResult` returned by that fn. Together they allow `WriteHandle::extend` (and by extension `WriteHandle::append`) to optimize the op-log by combining certain operations into one.

## Why
Reading/Watching about left-right/evmap it was mentioned that on top of the base doubling (unless carefully deduplicated) of memory the op-log can get quite out of hand if only infrequently published. That got me thinking about how, in a toy example of a simple map, you could combine set-ops with the same key to be the last one, merge certain modify-ops (like add), and completely consume the op-log when emitting a clear-op. Then I looked at the source and found it quite straightforward to implement. So I did (in about six hours).

## How
When compression is enabled `WriteHandle::extend` reverse-walks the fully unpublished portion of the op-log while attempting to combine ops. It does this by calling `Absorb::try_compress` and either:
 - skipping an empty entry
 - leaving behind an empty entry and continuing on when it returned `TryCompressResult::Compressed`
 - skipping an entry and continuing on when it returned `TryCompressResult::Independent`
 - stopping and inserting the op when it returned `TryCompressResult::Dependent`

After finishing for all ops it then stably moves all empty entries to the end of the op-log before truncating it.

I have also added three complementary optimizations to avoid some pathological worst cases:
 - It keeps track of the first non-empty entry from the back to avoid fully iterating a long but empty op-log in the case of a 'clear' (which consumes everything) being followed by a 'set'
 - It keeps track of its most recently encountered empty entry to avoid unnecessarily growing the op-log when inserting.
 - It stops after getting `Absorb::MAX_COMPRESS_RANGE` `TryCompressResult::Independent`s in a row to avoid iterating the full op-log in case it is filled with predominantly independent ops.

## Caveats
 - Even though they are only temporarily empty during `WriteHandle::extend`, the op-log now contains options instead of values. As ops are basically always enums, enum folding and `opt.unwrap_or_else(|| unreachable!())` should(?) remove any overhead.
 - Probably needs more tests than the ones I have already added.
 - Lots of slow but thorough `debug_assert!(...)`s, which should leave no traces in a release build.

## Is it useful?
Frankly, I don't know. I haven't actually used left-right in any serious projects and so can't really judge if this optimization is worthwhile. However, because compression is controlled via an associated const, it should monomorphize away and so at least not degrade performance when disabled, otherwise, it basically shifts writer-side computation from publishing to extending while potentially saving a fair chunk of memory. (Though come to think of it, it should also be quite straightforward to move the compression into `WriteHandle::publish`, though it then wouldn't be saving any memory, which likely is the main benefit.)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -189,6 +189,30 @@ pub use crate::read::{ReadGuard, ReadHandle, ReadHandleFactory};
 
 pub mod aliasing;
 
+/// The result of [`try_compress`](Absorb::try_compress).
+#[derive(Debug)]
+pub enum TryCompressResult<O> {
+    /// Returned when [`try_compress`](Absorb::try_compress) was successful.
+    Compressed {
+        /// The compressed op.
+        result: O,
+    },
+    /// Returned when [`try_compress`](Absorb::try_compress) failed because the ops are independent of each other.
+    Independent {
+        /// The op to be returned to the oplog then skipped.
+        prev: O,
+        /// The op to be inserted.
+        next: O,
+    },
+    /// Returned when [`try_compress`](Absorb::try_compress) failed because `prev` must precede `next`.
+    Dependent {
+        /// The op to be returned to the oplog preceding `next`.
+        prev: O,
+        /// The op to be inserted.
+        next: O,
+    },
+}
+
 /// Types that can incorporate operations of type `O`.
 ///
 /// This trait allows `left-right` to keep the two copies of the underlying data structure (see the
@@ -261,6 +285,29 @@ pub trait Absorb<O> {
     /// subtly affect results like the `RandomState` of a `HashMap` which can change iteration
     /// order.
     fn sync_with(&mut self, first: &Self);
+
+    /// Range at which [`WriteHandle`] tries to compress the oplog, reset each time a compression succeeds.
+    ///
+    /// Can be used to avoid having [`append`](WriteHandle::append) take O(oplog.len) if it is filled with mainly independent ops.
+    ///
+    /// Defaults to `&0`, which disables compression and allows the usage of an efficient fallback.
+    fn max_compress_range() -> &'static usize {
+        &0
+    }
+
+    /// Try to compress two ops into a single op to optimize the oplog.
+    ///
+    /// `prev` is the target of the compression and temporarily removed from the oplog, `next` is the op to be inserted.
+    ///
+    /// A return value of [`TryCompressResult::Compressed`] means the ops were successfully compressed,
+    /// [`TryCompressResult::Independent`] that while the ops can't be compressed,
+    /// `next` can safely precede `prev`, which therefore can be simply skipped over,
+    /// and [`TryCompressResult::Dependent`] that they can not be compressed, and `prev` must precede `next`.
+    ///
+    /// Defaults to `TryCompressResult::Dependent { prev, next }`, which disables compression.
+    fn try_compress(prev: O, next: O) -> TryCompressResult<O> {
+        TryCompressResult::Dependent { prev, next }
+    }
 }
 
 /// Construct a new write and read handle pair from an empty data structure.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -290,10 +290,8 @@ pub trait Absorb<O> {
     ///
     /// Can be used to avoid having [`append`](WriteHandle::append) take O(oplog.len) if it is filled with mainly independent ops.
     ///
-    /// Defaults to `&0`, which disables compression and allows the usage of an efficient fallback.
-    fn max_compress_range() -> &'static usize {
-        &0
-    }
+    /// Defaults to `0`, which disables compression and allows the usage of an efficient fallback.
+    const MAX_COMPRESS_RANGE: usize = 0;
 
     /// Try to compress two ops into a single op to optimize the oplog.
     ///
--- a/src/utilities.rs
+++ b/src/utilities.rs
@@ -8,13 +8,75 @@ impl Absorb<CounterAddOp> for i32 {
         *self += operation.0;
     }
 
-    fn absorb_second(&mut self, operation: CounterAddOp, _: &Self) {
-        *self += operation.0;
+    fn sync_with(&mut self, first: &Self) {
+        *self = *first
     }
+}
 
-    fn drop_first(self: Box<Self>) {}
+#[cfg(test)]
+#[derive(Debug)]
+pub enum CompressibleCounterOp<const MAX_COMPRESS_RANGE: usize> {
+    Set(i32),
+    Add(i32),
+    Sub(i32),
+}
+
+#[cfg(test)]
+impl<const MAX_COMPRESS_RANGE: usize> Absorb<CompressibleCounterOp<MAX_COMPRESS_RANGE>> for i32 {
+    fn absorb_first(
+        &mut self,
+        operation: &mut CompressibleCounterOp<MAX_COMPRESS_RANGE>,
+        _: &Self,
+    ) {
+        match operation {
+            CompressibleCounterOp::Set(v) => *self = *v,
+            CompressibleCounterOp::Add(v) => *self += *v,
+            CompressibleCounterOp::Sub(v) => *self -= *v,
+        }
+    }
 
     fn sync_with(&mut self, first: &Self) {
         *self = *first
+    }
+
+    fn max_compress_range() -> &'static usize {
+        &MAX_COMPRESS_RANGE
+    }
+
+    fn try_compress(
+        prev: CompressibleCounterOp<MAX_COMPRESS_RANGE>,
+        next: CompressibleCounterOp<MAX_COMPRESS_RANGE>,
+    ) -> TryCompressResult<CompressibleCounterOp<MAX_COMPRESS_RANGE>> {
+        match (prev, next) {
+            (CompressibleCounterOp::Add(prev), CompressibleCounterOp::Add(next)) => {
+                TryCompressResult::Compressed {
+                    result: CompressibleCounterOp::Add(prev + next),
+                }
+            }
+            (CompressibleCounterOp::Sub(prev), CompressibleCounterOp::Sub(next)) => {
+                TryCompressResult::Compressed {
+                    result: CompressibleCounterOp::Sub(prev + next),
+                }
+            }
+            (CompressibleCounterOp::Add(prev), CompressibleCounterOp::Sub(next)) => {
+                TryCompressResult::Independent {
+                    prev: CompressibleCounterOp::Add(prev),
+                    next: CompressibleCounterOp::Sub(next),
+                }
+            }
+            (CompressibleCounterOp::Sub(prev), CompressibleCounterOp::Add(next)) => {
+                TryCompressResult::Independent {
+                    prev: CompressibleCounterOp::Sub(prev),
+                    next: CompressibleCounterOp::Add(next),
+                }
+            }
+            (CompressibleCounterOp::Set(prev), next) => TryCompressResult::Dependent {
+                prev: CompressibleCounterOp::Set(prev),
+                next,
+            },
+            (_, CompressibleCounterOp::Set(next)) => TryCompressResult::Compressed {
+                result: CompressibleCounterOp::Set(next),
+            },
+        }
     }
 }
--- a/src/utilities.rs
+++ b/src/utilities.rs
@@ -39,9 +39,7 @@ impl<const MAX_COMPRESS_RANGE: usize> Absorb<CompressibleCounterOp<MAX_COMPRESS_
         *self = *first
     }
 
-    fn max_compress_range() -> &'static usize {
-        &MAX_COMPRESS_RANGE
-    }
+    const MAX_COMPRESS_RANGE: usize = MAX_COMPRESS_RANGE;
 
     fn try_compress(
         prev: CompressibleCounterOp<MAX_COMPRESS_RANGE>,
--- a/src/write.rs
+++ b/src/write.rs
@@ -484,7 +484,7 @@ where
             }
         } else {
             // Only try to compress if it could actually have an effect, else use original impl as efficient fallback
-            if *T::max_compress_range() > 0 {
+            if T::MAX_COMPRESS_RANGE > 0 {
                 // Compress oplog by rev-iterating all ops appended since the last publish
                 // while attempting to combine them with the next op,
                 // cut short when an attempt fails due to encountering a dependence (e.g. clear then set).
@@ -496,7 +496,7 @@ where
                     // used to more efficiently insert next if possible
                     let mut none: Option<(usize, &mut Option<O>)> = None;
                     // rev-iterate all unpublished ops already in the oplog
-                    let mut range_remaining = *T::max_compress_range();
+                    let mut range_remaining = T::MAX_COMPRESS_RANGE;
 
                     for (prev_rev_idx, prev_loc) in {
                         #[cfg(test)]
@@ -540,7 +540,7 @@ where
                                     // Remember empty loc for efficient insertion
                                     none.replace((prev_rev_idx, prev_loc));
                                     // We successfully compressed ops and therefore reset our range.
-                                    range_remaining = *T::max_compress_range();
+                                    range_remaining = T::MAX_COMPRESS_RANGE;
                                     next = result;
                                     // If the now empty loc is at the back of the non-none oplog we can increment none_back_count.
                                     if prev_rev_idx == none_back_count {
--- a/src/write.rs
+++ b/src/write.rs
@@ -28,7 +28,7 @@ where
 {
     epochs: crate::Epochs,
     w_handle: NonNull<T>,
-    oplog: VecDeque<O>,
+    oplog: VecDeque<Option<O>>,
     swap_index: usize,
     r_handle: ReadHandle<T>,
     last_epochs: Vec<usize>,
@@ -337,13 +337,20 @@ where
                 // we can drain out the operations that only the w_handle copy needs
                 //
                 // NOTE: the if above is because drain(0..0) would remove 0
-                for op in self.oplog.drain(0..self.swap_index) {
+                for op in self
+                    .oplog
+                    .drain(0..self.swap_index)
+                    .map(|op| op.unwrap_or_else(|| unreachable!("Nones are always temporary")))
+                {
                     T::absorb_second(w_handle, op, r_handle);
                 }
             }
             // we cannot give owned operations to absorb_first
             // since they'll also be needed by the r_handle copy
-            for op in self.oplog.iter_mut() {
+            for op in self.oplog.iter_mut().map(|op| {
+                op.as_mut()
+                    .unwrap_or_else(|| unreachable!("Nones are always temporary"))
+            }) {
                 T::absorb_first(w_handle, op, r_handle);
             }
             // the w_handle copy is about to become the r_handle, and can ignore the oplog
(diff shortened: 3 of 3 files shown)


jonhoo:
This is super interesting, and I'm not ignoring it, just need to find time to do a proper review!

Wulf0x67E7:
No problem. Gives me more time to iron out any kinks I find.

src/utilities.rs
```diff
@@ -8,13 +8,63 @@ impl Absorb<CounterAddOp> for i32 {
...
+                *prev += next;
+                TryCompressResult::Compressed
+            }
+            (CompressibleCounterOp::Sub(prev), CompressibleCounterOp::Sub(next)) => {
+                *prev += next;
+                TryCompressResult::Compressed
+            }
+            (CompressibleCounterOp::Add(_), CompressibleCounterOp::Sub(next)) => {
+                TryCompressResult::Independent(CompressibleCounterOp::Sub(next))
+            }
```
jonhoo, line 59:
```suggestion
            (CompressibleCounterOp::Add(_), next @ CompressibleCounterOp::Sub(_)) => {
                TryCompressResult::Independent(next)
            }
```
And similarly below.

src/write.rs
```diff
@@ -337,13 +337,20 @@ where
                 // we can drain out the operations that only the w_handle copy needs
                 //
                 // NOTE: the if above is because drain(0..0) would remove 0
-                for op in self.oplog.drain(0..self.swap_index) {
+                for op in self
+                    .oplog
+                    .drain(0..self.swap_index)
+                    .map(|op| op.unwrap_or_else(|| unreachable!("Nones are always temporary")))
```
jonhoo, line 343: Why `.unwrap_or_else(|| unreachable` here instead of just `.expect`?

Also, I think I'd prefer to do the unwrap/expect inside the loop instead of making the iterator this unwieldy.

src/write.rs
```diff
@@ -337,13 +337,20 @@ where
...
+                {
                     T::absorb_second(w_handle, op, r_handle);
                 }
             }
             // we cannot give owned operations to absorb_first
             // since they'll also be needed by the r_handle copy
-            for op in self.oplog.iter_mut() {
+            for op in self.oplog.iter_mut().map(|op| {
+                op.as_mut()
+                    .unwrap_or_else(|| unreachable!("Nones are always temporary"))
```
jonhoo, line 352: Same comment as above.

src/write.rs
```diff
@@ -476,7 +483,177 @@ where
                 Absorb::absorb_second(w_inner, op, &*r_handle);
             }
         } else {
-            self.oplog.extend(ops);
+            // Only try to compress if it is enabled, else use efficient fallback.
+            if T::MAX_COMPRESS_RANGE > 0 {
```
jonhoo, line 487: I'd like to invert the logic here.
```rust
if T::MAX_COMPRESS_RANGE == 0 {
    // .. easy case ..
    return;
}
// .. hard case ..
```

src/write.rs
```diff
@@ -476,7 +483,177 @@ where
...
+
+                // used to avoid walking more of the oplog than necessary.
+                let mut rev_dirty_range = 0usize..0;
+                for mut next in ops {
+                    // used to avoid linear insertion time in case of predominantly independent ops.
+                    let mut range_remaining = T::MAX_COMPRESS_RANGE;
+                    // used to more efficiently insert next if possible
+                    let mut none: Option<(usize, &mut Option<O>)> = None;
+                    // rev-iterate all unpublished and potentially non-none ops already in the oplog
+                    for (prev_rev_idx, prev_loc) in {
```
jonhoo, line 500: Instead of doing the sanity checking inline in the iterator, let's just have a separate `for` loop behind `#[cfg(test)]` above that checks all the relevant indicates, and then have _this_ `for` loop just do the non-test version.

src/write.rs
```diff
@@ -476,7 +483,177 @@ where
...
+                        #[cfg(test)]
+                        {
+                            // While testing, make very very sure we don't skip any Some.
+                            let none_rev_idx = rev_dirty_range.start.saturating_sub(1);
+                            self.oplog
+                                .iter_mut()
+                                .skip(self.swap_index) // only consider the fresh part of the oplog
```
jonhoo, line 507: I _believe_ `self.oplog[self.swap_index..].iter_mut()` is more efficient.

src/write.rs
```diff
@@ -476,7 +483,177 @@ where
...
+                                .rev() // We need to walk it in reverse
+                                .enumerate() // we need the reverse index for the rev_dirty_range optimization
+                                .skip_while(move |(rev_idx, loc)| {
+                                    if *rev_idx < none_rev_idx {
+                                        debug_assert!(loc.is_none(), "We never skip over Some.",);
```
jonhoo, line 512: This is already behind `#[cfg(test)]`, so can remain just an `assert`.

src/write.rs
jonhoo: Also note that `#[cfg(test)]` won't apply for integration tests, only unit tests. It may be that we want `#[cfg(debug_assertions)]` here instead.

src/write.rs
```diff
@@ -476,7 +483,177 @@ where
...
+                                        true
+                                    } else {
+                                        debug_assert!(
+                                            *rev_idx == none_rev_idx || loc.is_some(),
```
jonhoo, line 516: This means that we don't check whether `non_rev_idx` is a `None` or a `Some`. Is that intentional?

src/write.rs
```diff
@@ -476,7 +483,177 @@ where
...
-            self.oplog.extend(ops);
+            // Only try to compress if it is enabled, else use efficient fallback.
+            if T::MAX_COMPRESS_RANGE > 0 {
+                // Compress oplog by rev-iterating all ops appended since the last publish
+                // while attempting to combine them with the next op,
+                // cut short when an attempt fails due to encountering a dependency (e.g. clear then set).
+
+                // used to avoid walking more of the oplog than necessary.
+                let mut rev_dirty_range = 0usize..0;
+                for mut next in ops {
```
jonhoo, line 494: I wonder whether it might be more efficient to _first_ compress `ops` on its own, and _then_ compress the resulting list of operations into `self.ops`. What do you think?

jonhoo, line 494: Since we basically do the inserts one at a time here now, let's factor the entire contents of this loop into a dedicated function.

src/write.rs
```diff
@@ -476,7 +483,177 @@ where
...
+                        {
+                            self.oplog
+                                .iter_mut()
+                                .skip(self.swap_index) // only consider the fresh part of the oplog
+                                .rev() // We need to walk it in reverse
+                                .enumerate() // we need the reverse index for the rev_dirty_range optimization
+                                .skip(rev_dirty_range.start.saturating_sub(1)) // skip nones at the back (except one for efficient insertion)
+                        }
+                    } {
+                        if let Some(prev) = prev_loc.as_mut() {
```
jonhoo, line 533: Seems weird for it to be possible for this to be `None`. Can't that only happen if the oplog (after `self.swap_index`) is empty?

src/write.rs
```diff
@@ -476,7 +483,177 @@ where
...
+                            match T::try_compress(prev, next) {
+                                // The ops were successfully compressed, take prev as the new next
+                                crate::TryCompressResult::Compressed => {
+                                    // We successfully compressed ops and therefore take the combined op as the new next,...
+                                    next = prev_loc.take().unwrap_or_else(|| {
```
jonhoo, line 538: This should also use `.expect`

src/write.rs
```diff
@@ -476,7 +483,177 @@ where
...
+                    }
+                    // found nothing to combine with / encountered dependency
+                    // See if we found an empty loc during iteration, else push
+                    if let Some((none_rev_idx, none_loc)) = none {
+                        if none_loc.replace(next).is_some() {
+                            debug_assert!(
+                                false,
+                                "Should still have been None since we didn't insert yet."
+                            )
+                        }
```
jonhoo, line 588:
```suggestion
                        if let Some(_supposed_none) = none_loc.replace(next) {
                            unreachable!("cached None location held Some(_)");
                        }
```

src/write.rs
```diff
@@ -476,7 +483,177 @@ where
...
+                        if none_rev_idx + 1 == rev_dirty_range.end {
+                            rev_dirty_range.end = none_rev_idx;
+                        }
+                    } else {
+                        debug_assert!(rev_dirty_range.start == 0, "We only push if we found no nones and we always find at least one none at the back if it is there.");
+                        debug_assert!(rev_dirty_range.end == 0, "Pushing means no nones.");
+                        self.oplog.push_back(Some(next));
+                    }
+                }
+                // stably remove temporary nones from the oplog
```
jonhoo, line 603: I _think_ you can just `.sort_by_key(|o| o.is_none())` here, and then trim by the number of `None`s. That would save you much of the code below. You could probably also use `rev_dirty_range` to only sort the subslice that you know isn't all `None`s.

src/write.rs
```diff
@@ -576,6 +753,39 @@ mod tests {
         w.append(CounterAddOp(3));
         assert_eq!(w.oplog.len(), 2);
     }
+    #[test]
+    fn append_test_compress() {
```
jonhoo, line 757: I'd like to see more read assertions in this test, and not just oplog length assertions. Doesn't matter if the oplog is the right length if the value is wrong ;)

jonhoo (changes requested):
This is a really exciting change, and I love all the thought you've put into it. I think what I'd like to see above all is a benchmark to see that this compression actually manifests as a meaningful speedup to writes. I think there is an argument to be made that the compression logic may end up being slower than just _doing_ the extra operations, and would like to see some benchmarks refuting that. You'll probably want knobs to dictate the distance between publishes, the cost of execution an operation, and the cost of attempt to compress two operations.

Separately, this logic is fairly involved, and even though you've done a good job of writing test cases, I'd like to see a [`quickcheck`](https://github.com/BurntSushi/quickcheck) test or two added as well, just because there are _so_ many corner-cases to consider.

src/utilities.rs
```diff
@@ -8,13 +8,63 @@ impl Absorb<CounterAddOp> for i32 {
         *self += operation.0;
     }
 
-    fn absorb_second(&mut self, operation: CounterAddOp, _: &Self) {
-        *self += operation.0;
+    fn sync_with(&mut self, first: &Self) {
```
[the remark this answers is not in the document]
Wulf0x67E7: I think this was because it was the same as `absorb_first`, meaning it was identical to the default implementation, meaning this would have the tests also cover those. Not really important, can be re-added.

src/utilities.rs
Wulf0x67E7: Huh, didn't know you could do that. Neat.

src/write.rs
Wulf0x67E7: When I started learning Rust I remember hearing something about `.unwrap_or_else(|| unreachable!())` being more efficient and have just rolled with it since. Really thinking about it now, yeah, it really shouldn't be. Will change all occurrences.

src/write.rs
Wulf0x67E7: The compression working back-to-front instead of front-to-back means we would need to `collect` them first, meaning we would have to (de-)allocate every single call, even if only inserting a single op.

src/write.rs
Wulf0x67E7: Sadly VecDeque doesn't let us quickly sublice it. Would make everything so much nicer if we could. I looked at VecDeque's `nth`(and therefore `skip`) implementation though, and it uses an efficient index add-equal instead of looped increment. Should be fine.

src/write.rs
Wulf0x67E7: Yes. Because `let none_rev_idx = rev_dirty_range.start.saturating_sub(1);`, the first index we don't skip is None if there are any. It is used to avoid just pushing the op if compression fails.

src/write.rs
Wulf0x67E7: It is None if a previous iteration has left behind a gap.

[Some, None, Some, None, None] Would completely skip last None, start on and remember second-to-last None, try to compress with last Some, encounter and remember first None, try to compress with first Some

src/write.rs
Wulf0x67E7: `.sort_by_key(|o| o.is_none())` would be O(nLogn) instead of `.retain(|o| o.is_some()) with O(n), as well as possibly unstable. And because VecDeque doesn't let us sublice it, both would always be unnecessarily iterating the full oplog, even for only a single None at the back. Though I have effectively reimplemented `retain`, this version only iterates over the section of it that is actually mixed.

src/lib.rs
```diff
@@ -261,6 +272,32 @@ pub trait Absorb<O> {
...
+    /// Defaults to `0`, which disables compression and allows the usage of an efficient fallback.
+    const MAX_COMPRESS_RANGE: usize = 0;
+
+    /// Try to compress two ops into a single op to optimize the oplog.
+    ///
+    /// `prev` is the target of the compression and temporarily removed from the oplog, `next` is the op to be inserted.
+    ///
+    /// A return value of [`TryCompressResult::Compressed`] means the ops were successfully compressed,
+    /// [`TryCompressResult::Independent`] that while the ops can't be compressed,
+    /// `next` can safely precede `prev`, which therefore can be simply skipped over,
```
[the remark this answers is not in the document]
Wulf0x67E7: I mean that when [Add,Sub] <- Add and (Sub,Add) => Independent then we can skip Sub and try again with (Add,Add) => Compressed to be left with [Add2,Sub].

src/write.rs
Wulf0x67E7: Done: f49adcac65e0b22ac2130de393ec1df8f28f0c04

src/lib.rs
```diff
@@ -189,6 +189,17 @@ pub use crate::read::{ReadGuard, ReadHandle, ReadHandleFactory};
...
 pub mod aliasing;
 
+/// The result of calling [`Absorb::try_compress`](Absorb::try_compress).
+#[derive(Debug)]
+pub enum TryCompressResult<O> {
+    /// Returned when [`try_compress`](Absorb::try_compress) was successful.
+    Compressed,
+    /// Returned when [`try_compress`](Absorb::try_compress) failed because `prev` and `next` are independent of each other, contains `next`.
+    Independent(O),
+    /// Returned when [`try_compress`](Absorb::try_compress) failed because `prev` must precede `next`, contains `next`.
```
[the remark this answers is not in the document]
Wulf0x67E7: Done: cde672fe981acb4f1d5aa1e38c4dd4bc619a2ee2

src/lib.rs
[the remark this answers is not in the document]
Wulf0x67E7: Changed: cde672fe981acb4f1d5aa1e38c4dd4bc619a2ee2

src/lib.rs
```diff
@@ -261,6 +272,32 @@ pub trait Absorb<O> {
...
+    ///
+    /// `prev` is the target of the compression and temporarily removed from the oplog, `next` is the op to be inserted.
+    ///
+    /// A return value of [`TryCompressResult::Compressed`] means the ops were successfully compressed,
+    /// [`TryCompressResult::Independent`] that while the ops can't be compressed,
+    /// `next` can safely precede `prev`, which therefore can be simply skipped over,
+    /// and [`TryCompressResult::Dependent`] that they can not be compressed, and `prev` must precede `next`.
+    ///
+    /// Defaults to [`TryCompressResult::Dependent`], which sub-optimally disables compression.
+    /// Setting [`Self::MAX_COMPRESS_RANGE`](Absorb::MAX_COMPRESS_RANGE) to `0` is vastly more efficient for that.
```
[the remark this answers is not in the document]
Wulf0x67E7: Done: cde672fe981acb4f1d5aa1e38c4dd4bc619a2ee2

src/write.rs
Wulf0x67E7: Done: f74cadd688a8542f34d8e4797b37ae1c9cb920ae

src/write.rs
Wulf0x67E7: Done: f74cadd688a8542f34d8e4797b37ae1c9cb920ae

src/write.rs
Wulf0x67E7: Done: f74cadd688a8542f34d8e4797b37ae1c9cb920ae

src/write.rs
Wulf0x67E7: Done: 000d51264f3ac991359f656a0173eba70b3a3048

src/write.rs
Wulf0x67E7: Done: 03823329b34bf1f3503c9a0e082033ef86d22f3c

src/write.rs
Wulf0x67E7: Changed: 03823329b34bf1f3503c9a0e082033ef86d22f3c

src/write.rs
Wulf0x67E7: Done: cec99e0a119d6d07a427c34d041a226c55a437c3

src/write.rs
```diff
@@ -713,4 +971,170 @@ mod tests {
...
+        for x in input.iter_mut() {
+            *x &= 0b11;
+        }
+        type Op = CompressibleCounterOp<2>;
+        let (mut w, r) = crate::new::<i32, Op>();
+        // Get non-compressing first optimization out of the picture
+        w.publish();
+        assert_eq!(w.first, false);
+        // Map numbers to Ops, insert and publish them
+        let mut iter = input.chunks(8);
```
jonhoo, line 1107: Instead of manually setting the chunk size, let's allow quickcheck to set it to explore more corner-cases. If you take as input:
```rust
Vec<Vec<i32>>
```
that will give you chunks of varying lengths to then try to compress. If you want a little more control, try taking:
```rust
(chunk_sizes: Vec<u8>, input: Vec<i32>)
```
and then doing:
```rust
let chunk_sizes = chunk_sizes.into_iter().cycle();
let mut i = 0;
while i < input.len() {
    let chunk_size = chunk_sizes.next().unwrap().min(input.len() - i);
    let chunk = &input[i,i+chunk_size];
    // do the thing for the chunk
    i += chunk_size;
}
```
You could also change the `Vec<i32>` in `Vec<Vec<i32>>` to `(Vec<i32>, bool)` (or for the `Vec<u8>` chunking, change that to `Vec<(u8, bool)>`), and then use the `bool` to dictate whether or not you call `publish` after the `extend` for each chunk. That way you wouldn't need to hard-code the current "publish every other chunk".

src/write.rs
```diff
@@ -713,4 +971,170 @@ mod tests {
...
+                    } else {
+                        Op::Set(0)
+                    }
+                }));
+            }
+            w.publish();
+        }
+        // to avoid some macro complications
+        let val = *r.enter().unwrap();
+        // Check if value correct
```
jonhoo, line 1134: As you make the input logic more complex, I recommend computing the expected final value as you go, rather than trying to stitch it together from the input at the end. I suspect it'll ultimately lead to simpler code. But that's a matter of taste — either way is fine.

benches/benchmark.rs
```diff
@@ -0,0 +1,207 @@
+use std::collections::{BTreeMap, HashMap};
```
jonhoo, line 1: Could you post the results from these once you feel like they're in a decent place?

jonhoo (changes requested):
Took another look through — I'm really liking the changes so far!

src/lib.rs
```diff
@@ -261,6 +277,31 @@ pub trait Absorb<O> {
...
+    /// Try to compress two ops into a single op to optimize the oplog.
+    ///
+    /// `prev` is the target of the compression and temporarily removed from the oplog, `next` is the op to be inserted.
+    ///
+    /// A return value of [`TryCompressResult::Compressed`] means the ops were successfully compressed,
+    /// [`TryCompressResult::Independent`] that while the ops can't be compressed, `next` can safely precede `prev`,
+    /// and [`TryCompressResult::Dependent`] that they can not be compressed, and `prev` must precede `next`.
+    ///
+    /// Defaults to [`TryCompressResult::Dependent`], which sub-optimally disables compression.
+    /// Setting [`Self::MAX_COMPRESS_RANGE`](Absorb::MAX_COMPRESS_RANGE) to or leaving it at it's default of `0` is vastly more efficient for that.
```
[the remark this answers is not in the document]
Wulf0x67E7: I have vscode configured to automatically run `cargo fmt` on save and `cargo doc` doesn't complain about it, so I think it's fine? The single break between 'Defaults...' and 'Setting...' is purely meant to make reading it in an editor easier and should be turned into a simple space on docs.rs.

src/write.rs
```diff
@@ -476,7 +476,172 @@ where
                 Absorb::absorb_second(w_inner, op, &*r_handle);
             }
         } else {
-            self.oplog.extend(ops);
+            // Only try to compress if it is enabled, else use efficient fallback.
+            if T::MAX_COMPRESS_RANGE == 0 {
+                // efficient, non-compressing fallback
+                self.oplog.extend(ops.into_iter().map(|op| Some(op)));
```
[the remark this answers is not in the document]
Wulf0x67E7: Done: 1ca7bbc9f4bfeed800c02fea3dd37ba795ef325a

I instead combined the first optimization and the non-compressing variant into an else if. Same effect of removing one indentation, while also having all variants be on the same one.

src/write.rs
Wulf0x67E7: Done: b852b9ceb2600c78c8789f2a91ff46c4a4f876ec

benches/benchmark.rs
Wulf0x67E7: Will do. Done for today though. In their current state compression with unlimited range against no compression is about 20% faster for btreemap and about 35% faster for hashmap, with limited compression falling somewhere in between. Considering that due to the higher lookup-cost btreemap should be the one benefitting more there definitely is something fishy going on.

Wulf0x67E7:
While working on the benchmarks I just had an epiphany about an even more powerful and general improvement: Custom oplog data structures.

The original idea behind compression was to combine related ops, but the purely linear nature of the oplog meant I was essentially creating an unordered vec-map with linear lookup, which was fine and worthwhile for small maps but comes to a screeching halt for larger ones.

Then why not make the oplog itself a map and allow for custom append logic?

While this idea can be user-implemented by using compression to have a singleton op contain it (which is an excellent use case!), it would be less convoluted to have `Absorb` have an associated `OpLog` type and a `log_op(OpLog, Op)` method and either have `absorb_(first/second)` take the `OpLog` instead of single ops or have oplog have `(iter/drain)_ops -> Iterator`, the latter of which could be made to be backward-compatible but may require an additional `OpLog` trait and run into issues due to GATs.

I'm going to start another branch, tell me if you want me to open a new PR or merge it into this one.

Issue #101, opened 2022-03-17
Support for no_std?


Lol3rrr:
## Motivation
Im currently working on a Hobby project where this might be very useful, however the Project is written for a no_std environment meaning that I currently can not use this Crate.

## Situation
From quickly looking through the Code I found that most of the uses of `std` can be easily replaced by either `core` or `alloc`, expect for the use of `std::thread::yield_now` in [this line](https://github.com/jonhoo/left-right/blob/d6d13125af0b02f44bc299329d56773c2b913c7e/src/write.rs#L278) for the WriteHandle

### Boxed yield closure
We could store a boxed closure for yield in the Write Handle, which would by default just call `std::thread::yield_now` but could then also be changed to support other mechanisms, which might also enable an async refresh where we dont block the thread but rather yield the current future.

However this would also introduce the extra Cost of another Allocation + calling the closure through the Box for everyone, although the writer is usually not that important for speed in this crate I would rather avoid unnecessary overhead where possible.

### Generic-Type-Parameter yield closure
This idea is basically the same as the Boxed yield closure one, although instead of boxing a closure we would instead store the closure or function-pointer in a field and have its type specified using a Generic-Type-Parameter, which would avoid the Cost of the Box + overhead. However this would increase the overall Complexity of the WriteHandle Type, which might not be desireable. However this also gives the most flexability for consumers of the library.

### Function Pointers
This idea is again like the previous 2, but we now store a simple function-pointer instead of a closure. This would mitigate both the Problem of extra heap allocations and also not increase the complexity of the WriteHandle Type. However this might limit the consumers of the Library as the function has to be stateless, but this might not matter much as the current function is also stateless.

These were just my initial thoughts on this and I would love to hear others thoughts about it as well. If we come to a conclusion everyone is happy with, I would also happily make these changes and implement it

Lol3rrr:
After looking at the Code more closely, I also found that it uses the `std::Mutex` for epochs and I dont really know how we can adjust that without adding a new dependency like [spin](https://crates.io/crates/spin) which I would like to avoid for now. Any ideas would be welcome

jonhoo:
[...]

PR #102, opened 2022-03-19
The Pull Request for Issue 101 for no_std

Lol3rrr:
This is the corresponding for Issue #101.

Currently this only contains the changes for the yielding of the WriteHandle

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -277,6 +277,18 @@ where
     (w, r)
 }
 
+/// TODO
+pub fn new_from_empty_with_yield<T, O>(t: T, yield_fn: fn()) -> (WriteHandle<T, O>, ReadHandle<T>)
+where
+    T: Absorb<O> + Clone,
+{
+    let epochs = Default::default();
+
+    let r = ReadHandle::new(t.clone(), Arc::clone(&epochs));
+    let w = WriteHandle::new_with_yield(t, epochs, r.clone(), yield_fn);
+    (w, r)
+}
+
 /// Construct a new write and read handle pair from the data structure default.
 ///
 /// The type must implement `Default` so we can construct two empty instances. You must ensure that
@@ -295,5 +307,17 @@ where
 
     let r = ReadHandle::new(T::default(), Arc::clone(&epochs));
     let w = WriteHandle::new(T::default(), epochs, r.clone());
+    (w, r)
+}
+
+/// TODO
+pub fn new_with_yield<T, O>(yield_fn: fn()) -> (WriteHandle<T, O>, ReadHandle<T>)
+where
+    T: Absorb<O> + Default,
+{
+    let epochs = Default::default();
+
+    let r = ReadHandle::new(T::default(), Arc::clone(&epochs));
+    let w = WriteHandle::new_with_yield(T::default(), epochs, r.clone(), yield_fn);
     (w, r)
 }
--- a/src/write.rs
+++ b/src/write.rs
@@ -3,12 +3,12 @@ use crate::Absorb;
 
 use crate::sync::{fence, Arc, AtomicUsize, MutexGuard, Ordering};
 use std::collections::VecDeque;
+use std::fmt;
 use std::marker::PhantomData;
 use std::ops::DerefMut;
 use std::ptr::NonNull;
 #[cfg(test)]
 use std::sync::atomic::AtomicBool;
-use std::{fmt, thread};
 
 /// A writer handle to a left-right guarded data structure.
 ///
@@ -42,6 +42,9 @@ where
     second: bool,
     /// If we call `Self::take` the drop needs to be different.
     taken: bool,
+    /// This function will be used to yield the current execution instead of just spinning while
+    /// waiting for all readers to move on
+    yield_fn: fn(),
 }
 
 // safety: if a `WriteHandle` is sent across a thread boundary, we need to be able to take
@@ -213,6 +216,15 @@ where
     T: Absorb<O>,
 {
     pub(crate) fn new(w_handle: T, epochs: crate::Epochs, r_handle: ReadHandle<T>) -> Self {
+        Self::new_with_yield(w_handle, epochs, r_handle, std::thread::yield_now)
+    }
+
+    pub(crate) fn new_with_yield(
+        w_handle: T,
+        epochs: crate::Epochs,
+        r_handle: ReadHandle<T>,
+        yield_fn: fn(),
+    ) -> Self {
         Self {
             epochs,
             // safety: Box<T> is not null and covariant.
@@ -228,6 +240,7 @@ where
             first: true,
             second: true,
             taken: false,
+            yield_fn,
         }
     }
 
@@ -275,7 +288,7 @@ where
                         if iter != 20 {
                             iter += 1;
                         } else {
-                            thread::yield_now();
+                            (self.yield_fn)();
                         }
                     }
 


src/lib.rs
```diff
@@ -1,3 +1,4 @@
+#![cfg_attr(not(feature = "std"), no_std)]
```
jonhoo, line 1: Let's move this down to below the crate-level doc comment.

src/lib.rs
```diff
@@ -263,20 +268,47 @@ pub trait Absorb<O> {
     fn sync_with(&mut self, first: &Self);
 }
 
+/// Construct a new write and read handle pair from an empty data structure with
+/// [yield_now](std::thread::yield_now) as the yield function.
```
jonhoo, line 272:
```suggestion
/// [`yield_now`](std::thread::yield_now) as the yield function.
```

src/lib.rs
```diff
@@ -263,20 +268,47 @@ pub trait Absorb<O> {
...
+///
+/// See [new_from_empty_with_yield] for a more detailed explaination
```
jonhoo, line 274:
```suggestion
/// See [`new_from_empty_with_yield`] for a more detailed explanation.
```

src/lib.rs
```diff
@@ -263,20 +268,47 @@ pub trait Absorb<O> {
...
     let epochs = Default::default();
 
     let r = ReadHandle::new(t.clone(), Arc::clone(&epochs));
-    let w = WriteHandle::new(t, epochs, r.clone());
+    let w = WriteHandle::new_with_yield(t, epochs, r.clone(), yield_fn);
     (w, r)
 }
 
+/// Construct a new write and read handle pair from the data structure default, with
+/// [yield_now](std::thread::yield_now) as the yield function.
```
jonhoo, line 301:
```suggestion
/// [`yield_now`](std::thread::yield_now) as the yield function.
```

src/lib.rs
```diff
@@ -263,20 +268,47 @@ pub trait Absorb<O> {
...
+///
+/// See [new_with_yield] for a more detailed explaination
```
jonhoo, line 303:
```suggestion
/// See [`new_with_yield`] for a more detailed explanation.
```

src/sync.rs
```diff
@@ -17,6 +17,8 @@ pub(crate) fn fence(ord: Ordering) {
 }
 
 #[cfg(not(loom))]
-pub(crate) use std::sync::atomic::{fence, AtomicPtr, AtomicUsize, Ordering};
+pub(crate) use alloc::sync::Arc;
 #[cfg(not(loom))]
-pub(crate) use std::sync::{Arc, Mutex, MutexGuard};
+pub(crate) use core::sync::atomic::{fence, AtomicPtr, AtomicUsize, Ordering};
+#[cfg(not(loom))]
+pub(crate) use std::sync::{Mutex, MutexGuard};
```
jonhoo, line 24: Just a note to self that this is still a `std` dependency as we've discussed in #101.

jonhoo (commented):
Overall this looks good. I left some minor nits, plus a TODO for the `Mutex` bit. We'll also want to update CI to test compiling in a no_std environment. You can use [this CI file](https://github.com/jonhoo/haphazard/blob/main/.github/workflows/nostd.yml) as a starting point.

Lol3rrr:
Fixed all the Notes from the Review and now also added the Lock-Free list, called HandleList.
Also updated most of the Tests to support testing in no_std where possible, however this is not possible for the Doc-Test in the main comment so unsure if this should even be touched or just ignored because all other tests work in no_std.

Otherwise it should now be fully no_std compliant as far as I can tell and just needs some minor finishing touches in terms of internal comments and style in certain places

jonhoo:
Hmm, I think this is going to end up leaking all the entries since there's no `Drop` implementation. I wonder if the trick here is going to be using some existing crate that provides a lock free list — what do you think?

Lol3rrr:
While using an existing crate for a lock free list would certainly make it a bit easier since we dont need to maintain our own implementation. However I dont know how we could then have these "snapshots" for the write handle, because the crate would need to support it or something similiar.

I think for now I might try to improve the List I wrote and see if I can improve it a bit, but if that does not work out we should consider using some other crate.

jonhoo:
I _think_ basically any singly-linked atomic linked list gives you "snapshot" read iterators. They sort of have to by nature — if you only add to the head, then any read has to start at the head at the time of the first read, and just walk "backwards in time" from then.

Lol3rrr:
Yeah right, didnt really think about that, but I think there was a Problem, where want to iterate over the same list of handles in the Writer, while we wait for them to move on ([new line](https://github.com/Lol3rrr/left-right/blob/72ae2c6b540835d28d0cffddf7d4267c9d059aa4/src/write.rs#L259) where we create a new Iterator in a loop vs. [old line](https://github.com/jonhoo/left-right/blob/d6d13125af0b02f44bc299329d56773c2b913c7e/src/write.rs#L246) where we create a new Iterator in the loop but based on the MutexGuard so its always the same).
Thats why I create an explicit snapshot and then create an iterator based of that snapshot, because we will then always iterate over the same Handles.

If we were to just create a new Iterator for every time where we check them, there might be a new Handle added and we need to keep track of which handles we saw last time and where they are now (mapping between current list state and last seen "sub-list")

jonhoo:
That's a good point. However, I _think_ you'll generally find the iterators to be `Clone` specifically for this reason. When you grab an iterator for a linked list, it really just holds a pointer to a head node. And for an shared reference iterator, that iterator _should_ be `Clone` so that you can iterate from that same node multiple times.

src/write.rs
```diff
@@ -1,14 +1,17 @@
...
-use crate::sync::{fence, Arc, AtomicUsize, MutexGuard, Ordering};
-use std::collections::VecDeque;
-use std::marker::PhantomData;
-use std::ops::DerefMut;
-use std::ptr::NonNull;
 #[cfg(test)]
-use std::sync::atomic::AtomicBool;
-use std::{fmt, thread};
+use crate::sync::Arc;
+use crate::sync::{fence, Ordering};
```
jonhoo, line 7: Let's combine these.

src/read.rs
```diff
@@ -53,11 +53,8 @@ unsafe impl<T> Send for ReadHandle<T> where T: Sync {}
 
 impl<T> Drop for ReadHandle<T> {
     fn drop(&mut self) {
-        // epoch must already be even for us to have &mut self,
-        // so okay to lock since we're not holding up the epoch anyway.
-        let e = self.epochs.lock().unwrap().remove(self.epoch_i);
-        assert!(Arc::ptr_eq(&e, &self.epoch));
-        assert_eq!(self.enters.get(), 0);
+        // We dont need a Drop implementation as of now, because the Epoch for this Handle will not
+        // be freed again
```
jonhoo, line 57: I don't think it's okay for us to only drop a handle only after _all_ handles have been dropped. Doing so would effectively mean that the application has unbounded memory growth, and if it for example creates read a new read handle regularly or in a loop or something, it'll run into problems. I think we _have_ to support removal, even if it's maybe delayed or periodic.

src/write.rs
```diff
@@ -168,9 +174,11 @@ where
         let r_handle = self.r_handle.inner.swap(ptr::null_mut(), Ordering::Release);
 
         // now, wait for all readers to depart
-        let epochs = Arc::clone(&self.epochs);
-        let mut epochs = epochs.lock().unwrap();
-        self.wait(&mut epochs);
+        //let epochs = Arc::clone(&self.epochs);
+        //let mut epochs = epochs.lock().unwrap();
+
```
jonhoo, line 179:
```suggestion
```

src/write.rs
```diff
@@ -228,22 +241,24 @@ where
...
 
         #[cfg(test)]
         {
             self.is_waiting.store(true, Ordering::Relaxed);
         }
-        // we're over-estimating here, but slab doesn't expose its max index
-        self.last_epochs.resize(epochs.capacity(), 0);
+
+        // make sure we have enough space for all the epochs in the current snapshot
+        self.last_epochs.resize(epochs.iter().count(), 0);
```
jonhoo, line 258: It's a little unfortunate to end up having to traverse the list twice. It'd be better if we could avoid that.

src/write.rs
```diff
@@ -303,14 +318,10 @@ where
...
         // readers present since we did the previous refresh)
-        //
-        // NOTE: it is safe for us to hold the lock for the entire duration of the swap. we will
-        // only block on pre-existing readers, and they are never waiting to push onto epochs
-        // unless they have finished reading.
-        let epochs = Arc::clone(&self.epochs);
-        let mut epochs = epochs.lock().unwrap();
 
-        self.wait(&mut epochs);
+        let epoch_snapshot = self.epochs.snapshot();
```
jonhoo, line 322: I think the NOTE that was removed here still applies, it's just about "not seeing new readers" rather than "holding the lock".

src/handle_list.rs
```diff
@@ -0,0 +1,227 @@
...
+    fn clone(&self) -> Self {
+        Self {
+            inner: Arc::clone(&self.inner),
+        }
+    }
+}
+
+impl ListSnapshot {
+    /// Obtain an iterator over the Entries in this Snapshot
+    pub fn iter(&self) -> SnapshotIter {
```
jonhoo, line 134: This isn't safe. `SnapshotIter` will need a lifetime argument that's tied to the lifetime of `&self` here to ensure that someone doesn't write code like this:
```rust
let x = HandleList::new();
let _ = x.new_entry();
let mut snapshot = x.snapshot();
drop(x);
let oops = snapshot.next().unwrap();
```

src/handle_list.rs
```diff
@@ -0,0 +1,227 @@
...
+        SnapshotIter { current: self.head }
+    }
+}
+
+impl Iterator for SnapshotIter {
+    // TODO
+    // Maybe don't return an owned Value here
+    type Item = Arc<AtomicUsize>;
```
jonhoo, line 142: Once `SnapshotIter` becomes `SnapshotIter<'list>`, this can become
```rust
type Item = &'list Arc<AtomicUsize>;
```

src/handle_list.rs
```diff
@@ -0,0 +1,227 @@
...
+
+    fn next(&mut self) -> Option<Self::Item> {
+        if self.current.is_null() {
+            return None;
+        }
+
+        // Safety
+        // The Ptr is not null, because of the previous if-statement.
+        // The Data is also not freed, because we never free Entries on the List.
```
jonhoo, line 151: This is no longer true.

src/handle_list.rs
```diff
@@ -0,0 +1,227 @@
...
+        // We also have no one mutating Entries on the List and therefore we can access this without
+        // any extra synchronization needed.
```
jonhoo, line 153: This will no longer be true once we have removal of individual elements.

src/handle_list.rs
```diff
@@ -0,0 +1,227 @@
...
+        self.current = entry.next;
+
+        Some(entry.data.clone())
+    }
+}
+
+impl Drop for InnerList {
+    fn drop(&mut self) {
+        // We iterate over all the Entries of the List and free every Entry of the List
+        let mut current = self.head.load(Ordering::SeqCst);
```
jonhoo, line 165: I think you can use `self.head.get_mut()` here since we have `&mut self`.

src/handle_list.rs
```diff
@@ -0,0 +1,227 @@
...
+        while !current.is_null() {
+            // # Safety
+            // This is safe, because we only enter the loop body if the Pointer is not null and we
+            // also know that the Entry is not yet freed because we only free them once we are dropped
+            // and because we are now in Drop, noone before us has freed any Entry on the List
+            let current_r = unsafe { &*current };
+
+            let next = current_r.next as *mut ListEntry;
```
jonhoo, line 173:
```suggestion
            let next = unsafe { &*current }.next as *mut ListEntry;
```
This is better so that we don't leave `current_r` around after we've run `Box::from_raw`, since accessing it then would be undefined behavior.

src/handle_list.rs
```diff
@@ -0,0 +1,227 @@
...
+
+            // # Safety
+            // This is safe, because of the same garantuees detailed above for `current_r`
```
jonhoo, line 176: The safety requirements aren't quite the same for `&*` and `Box::from_raw`. First, we have to guarantee that the pointer was originally created using `Box::into_raw` (which we know it was). And second, we need to ensure that _no one_ else has, or ever will have, a reference to the value behind the pointer (basically, we have to guarantee that it's not aliased). That we know because we take `&mut self` here, and we know that all references to any node in the list is tied to `InnerList`.

jonhoo (changes requested):
Thanks for checking back in on this! I left some more in-depth notes this time.

src/read.rs
Lol3rrr: Yeah I also thought about this but we would also run in the Problem of having to deal with memory reclaimation. What would you think about not actually removing the entries but instead be able to reuse entries in the List, I haven't thought about it too much yet and it would not entirely solve the Problem with unbounded memory usage but It would at least help mitigate most cases. Just something to maybe consider as well

jonhoo: Ah, yes, reuse would be a nice way to go about it!

src/handle_list.rs
Lol3rrr: Thats a good catch, thanks. Just fixed it

src/handle_list.rs
Lol3rrr: Already switched although not to `&'list Arc<AtomicUsize>` but instead to `&'list AtomicUsize` to avoid an extra indirection through the Arc.

src/handle_list.rs
Lol3rrr: Updated it with the correct reasoning now, though not quite sure on the wording I used

src/handle_list.rs
Lol3rrr: Not really, because we just need the actual Pointer stored in the head. So for now I still replaced it with `*self.head.get_mut()` but I'm not really sure if this actually better than the load.

src/write.rs
Lol3rrr: I tried to update the Note to reflect this now, not quite sure on the wording there though

Lol3rrr:
Fixed most of the Comments you made for the last commit.

Although there are still some open questions for me, especially how we would want to do the removal of entries in the List, which I already touched on in one of my comments. My main concern is that we would need to either have nodes be able to represent an "empty" state, which would certainly make it more complicated and would mean that we would never be able to actually reduce our memory usage. However if we start removing entries we would need to worry about memory reclaimation, which Im sure you know can be very tricky (especially when targeting no_std i think).

Besides that there is only the small question of how we could avoid having to traverse the List twice in `publish`, but thats also not of the highest priority compared to the other problem we need to solve

jonhoo:
I think having a way to represent "unused" is the way to go here. We don't actually have to _empty_ them, since the whole `Arc<AtomicUsize>` can be re-used by another reader that comes along later! The "used" bit can just be an `AtomicBool` that you `compare_exchange` from `false` to `true`, and keep searching (or eventually allocate) if nothing is found.

You may find this code from the `haphazard` crate which more or less implements the exact same thing helpful:

 - Try to grab an unused node: https://github.com/jonhoo/haphazard/blob/e0e18f60f78652a63aba235be854f87d106c1a1b/src/domain.rs#L277
 - Allocate a new node if no unused are available: https://github.com/jonhoo/haphazard/blob/e0e18f60f78652a63aba235be854f87d106c1a1b/src/domain.rs#L388

Maybe we could even take that implementation and extract it into its own crate so we can re-use it!

jonhoo:
I extracted that list into its own crate: [`sento`](https://github.com/jonhoo/sento). Give that a try?

Lol3rrr:
Hopefully I can find some time this week to properly look into it and integrate that into it. Will give an update then

indietyp:
I would love to use left-right in a no-std project of mine, but development on no-std support seems to have stalled, what is the current state of development and is there any way I could help out?

Lol3rrr:
Im not exactly sure anymore what exactly the stopping factor was, but from a quick look there seem to be two points that need to be addressed:
- Add CI test for the no_std implementation
- Switch to the proposed [sento](https://github.com/jonhoo/sento) crate for the List

PR #109, opened 2022-08-31
add a test with a VecDeque structure

RaduBerinde:
Hi @jonhoo - I watched your stream (thanks for all the content!) and wanted to get some hands-on experience with this data structure. I wrote a test that implements a deque with aliased values.

This commit adds a test that uses the left-right primitive with a
VecDeque and checks for correct dropping of values.

- - -
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/left-right/109)

### DIFF
--- a/tests/deque.rs
+++ b/tests/deque.rs
@@ -1,0 +1,162 @@
+use std::cell::Cell;
+use std::collections::VecDeque;
+use std::rc::Rc;
+
+use left_right::{
+    aliasing::{Aliased, DropBehavior},
+    Absorb, ReadHandle,
+};
+
+// Value encapsulates an integer value and keeps a registry of live values up to
+// date.
+struct Value {
+    v: i32,
+    r: Rc<ValueRegistry>,
+}
+
+impl Value {
+    fn new(v: i32, r: Rc<ValueRegistry>) -> Self {
+        r.adjust_count(1);
+        Self { v, r }
+    }
+}
+
+impl Drop for Value {
+    fn drop(&mut self) {
+        self.r.adjust_count(-1);
+    }
+}
+
+// ValueRegistry keeps track of the number of Values that have been created and
+// not yet dropped.
+struct ValueRegistry {
+    num_live_values: Cell<i64>,
+}
+
+impl ValueRegistry {
+    fn new() -> Self {
+        Self {
+            num_live_values: Cell::new(0),
+        }
+    }
+
+    fn adjust_count(&self, delta: i64) {
+        let mut live_vals = self.num_live_values.get();
+        live_vals += delta;
+        assert!(live_vals >= 0);
+        self.num_live_values.set(live_vals);
+    }
+
+    fn expect(&self, expected_count: i64) {
+        assert_eq!(self.num_live_values.get(), expected_count);
+    }
+}
+
+struct NoDrop;
+impl DropBehavior for NoDrop {
+    const DO_DROP: bool = false;
+}
+
+struct DoDrop;
+impl DropBehavior for DoDrop {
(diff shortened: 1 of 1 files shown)


tests/deque.rs
```diff
@@ -0,0 +1,158 @@
...
+        *live_vals += delta;
+        assert!(*live_vals >= 0);
+    }
+
+    fn expect(&self, expected_count: i64) {
+        assert_eq!(*self.num_live_values.borrow(), expected_count);
+    }
+}
+
+struct NoDrop;
```
RaduBerinde, line 54: It feels strange that every user has to implement these types. Wouldn't it be much simpler to just use a bool const instead of a type? (`pub struct Aliased<T, const DoDrop: bool>`).

RaduBerinde: I guess the idea is that someone could implement a trait only for `Aliased<T, true>` potentially (in theory) causing the memory layout to change for just the `true` case?

tests/deque.rs
```diff
@@ -0,0 +1,158 @@
...
+// ValueRegistry keeps track of the number of Values that have been created and
+// not yet dropped.
+struct ValueRegistry {
+    num_live_values: RefCell<i64>,
+}
+
+impl ValueRegistry {
+    fn new() -> Self {
+        Self {
+            num_live_values: RefCell::new(0),
```
jonhoo, line 39: You could just use `Cell` here rather than `RefCell` since the value doesn't need to be `Sync` :)

tests/deque.rs
jonhoo: The reason why these _have_ to be types private to each implementor is explained in the "Unsafe casting" section [of the `mod aliasing` docs](https://docs.rs/left-right/latest/left_right/aliasing/index.html#unsafe-casting). Basically, it's the only way you can guarantee that the necessary cast is actually sound.

tests/deque.rs
```diff
@@ -0,0 +1,158 @@
...
+                self.push_back(unsafe { value.alias() });
+            }
+            Op::PopFront => {
+                self.pop_front();
+            }
+        }
+    }
+
+    fn absorb_second(&mut self, operation: Op, _other: &Self) {
+        let with_drop: &mut VecDeque<Aliased<Value, DoDrop>> =
```
jonhoo, line 83: I'd like to see a comment here explaining why this is safe (I agree that it is).

tests/deque.rs
```diff
@@ -0,0 +1,158 @@
...
+        assert_eq!(self.len(), 0);
+        self.extend(first.iter().map(|v| unsafe { v.alias() }));
+    }
+
+    fn drop_first(self: Box<Self>) {
+        // The Deque type has NoDrop, so this will not drop any of the values.
+    }
+
+    fn drop_second(self: Box<Self>) {
+        // Comnvert self to DoDrop and drop it.
```
jonhoo, line 105:
```suggestion
        // Convert self to DoDrop and drop it.
```

tests/deque.rs
```diff
@@ -0,0 +1,158 @@
...
+        let with_drop: Box<VecDeque<Aliased<Value, DoDrop>>> =
+            unsafe { Box::from_raw(Box::into_raw(self) as *mut _ as *mut _) };
```
jonhoo, line 107: Same here: I'd like a comment explaining why this is safe

jonhoo (changes requested):
This is great! Thanks for taking the time. I left some comments inline.

RaduBerinde:
Thanks for the review! Updated.

jonhoo (approved):
Thanks!

PR #119, opened 2025-08-10
Refactor publish to remove ‘first’ and ‘second’ edge cases

manuelgdlvh:
**Description**

This PR updates the publish method implementation to no longer treat the first and second publish operations as special cases within the base algorithm. These operations were handled directly rather than being passed through the operation log, introducing conditional branches into the base algorithm.

While this special handling could reduce overhead for initial operations, it added complexity and required synchronization via a dedicated API method. From an API implementor’s perspective, this makes the implementation slightly trickier—ideally, a simple clone() during initialization should be sufficient.

At this point, there is no performance benchmark to validate whether the previous optimization (handling the first two publishes separately) offers meaningful gains. I plan to add benchmarks to compare both approaches and evaluate whether the added complexity is justified.

This change is [Reviewable](https://reviewable.io/reviews/jonhoo/left-right/119)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -94,9 +94,6 @@
 //!     // See the documentation of `Absorb::drop_first`.
 //!     fn drop_first(self: Box<Self>) {}
 //!
-//!     fn sync_with(&mut self, first: &Self) {
-//!         *self = *first
-//!     }
 //! }
 //!
 //! // Now, you can construct a new left-right over an instance of your data structure.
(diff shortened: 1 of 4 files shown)


jonhoo:
The `first` and `second` distinction isn't just for initial operations, it's to allow implementations to handle aliased values between the two maps correctly. For example, if you alias a `Box`ed value between the two halves, it's only safe to drop the re-constituted `Box` in `second`.

PR #138, opened 2026-03-06
Add an accumulator type to the Absorb trait

rs-sac:
When applying several operations to the same value, certain operations may be faster to compute once at the end rather than incrementally after each change.  E.g. if values are `Vec<u64>`, it would be faster to compute an average or to sort once, after all updates are applied, rather than incrementally after each update.

Add an associated type to the trait for the accumulator, pass it to the absorb methods, and call a new finalize method at the end of each absorb loop.  If not needed, the accumulator can be set to `()` and the finalize method can be left with its default no-op implementation.

- - -
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/left-right/138)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -61,7 +61,7 @@ checksum = "bbd2bcb4c963f2ddae06a2efc7e9f3591312473c50c6685e1f298068316e66fe"
 
 [[package]]
 name = "left-right"
-version = "0.11.7"
+version = "0.12.0"
 dependencies = [
  "crossbeam-utils",
  "loom",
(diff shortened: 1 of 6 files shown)


jonhoo:
@rs-sac After we discussed this offline, do you still think you need this?

For anyone reading this, the quick summary was to change the `Op` type to instead be "batches" of operations, which would then allow you do this special handling directly in `absorb`, rather than needing to pass state _between_ `absorb` calls.

rs-sac:
We don't need this for now.  The jury may still be a little out, but I implemented batching and merging outside of left-right.

PR #143, opened 2026-04-05
Don't say that `Absorb` is unsafe

tyilo:
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/left-right/143)


jonhoo (approved):
Ooops, thanks for catching this!

PR #144, opened 2026-04-24
Fix data race on writehandle drop

Fredi-raspall:
Fix writehandle drop logic that allows readhandles to enter() while the value is being dropped.

The issue was observed by building with thread sanitizer in a test where:
* one thread T1  creates some `writehandle<V> `and shares a readhandle factory for it with some other threads
* T1 continuously creates and drops `writehandle<V> `(sharing the factory accordingly), while the other threads create new readhandles from the factory and attempt to `enter()`.
* TSAN reports race where some threads manage to read a bool within `V` while T1 drops it.

### DIFF
--- a/src/write.rs
+++ b/src/write.rs
@@ -172,10 +172,23 @@ where
         // now, wait for all readers to depart
         let epochs = Arc::clone(&self.epochs);
         let mut epochs = epochs.lock().unwrap();
-        self.wait(&mut epochs);
 
         // ensure that the subsequent epoch reads aren't re-ordered to before the swap
         fence(Ordering::SeqCst);
+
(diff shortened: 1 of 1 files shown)


Fredi-raspall:
Hello! I opened this a couple of weeks ago, but have not got any feedback. No rush, I just wanted to know whether the issue is clear, the fix makes sense at all, or something else is needed. Thanks!

jonhoo:
Sorry for the delay — was in the middle of a house move, so time has been tight! I wanted to make some smaller modifications to the comments here, but it seems I can't push to your branch, so I'm landing this in https://github.com/jonhoo/left-right/pull/153. There was definitely a bug here :)

One request though: could you open a PR where you add the TSAN test case so that we catch regressions on this in the future?

PR #145, opened 2026-04-27
Replace `Mutex` with `cordyceps::TransferStack` for `Epochs` and make no_std compatible

ilslv:
Hi! I've recently saw a [talk](https://www.youtube.com/watch?v=tND-wBBZ8RY) and was wondering what will it take to remove `Mutex` lock on `ReadHandle` initialisation and drop.

The simplest approach I could come boils down to the following idea: what if `ReadHandle`s wouldn't bother themselves with removal of their `Epoch`s from shared `Epochs` and would just drop the `Arc`. Then `WriteHandle` can check for `ReadHandle` liveliness with `Arc::strong_count` since every `Arc` has a strong count of 1 or 2. After that it became obvious that something like `cordyceps::TransferStack` can be used to semantically replace `Epochs` from "shared storage of `Epoch`s that is accessed both by `ReadHandle`s and `WriteHandle`" to "a kinda channel to let `WriteHandle` know about new `ReadHandle`s".

I also made library `no_std` compatible as mentioned in https://github.com/jonhoo/left-right/issues/101.

I think more tests should be added before merging, but before investing more time I would love to here about approach in general and haven't I missed something.


Cargo.toml
```diff
@@ -11,8 +11,12 @@ repository = "https://github.com/jonhoo/left-right.git"
 keywords = ["concurrency","lock-free"]
 categories = ["concurrency"]
 
+[features]
+default = ["std"]
+std = ["cordyceps/std"]
+
 [dependencies]
-slab = "0.4.1"
+cordyceps = { version = "0.3" }
```
ilslv: `cordyceps` has a following note 

> This is a hobby project. I’m working on it in my spare time, for my own personal use. I’m very happy to share it with the broader Rust community, and [contributions](https://github.com/hawkw/mycelium/compare) and [bug reports](https://github.com/hawkw/mycelium/issues/new) are always welcome. However, please remember that I’m working on this library for fun, and if it stops being fun…well, you get the idea.
>
> Anyway, feel free to use and enjoy this crate, and to contribute back as much as you want to!

But implementation looks very solid and [`embassy-executor`](https://crates.io/crates/embassy-executor) uses it

src/epochs.rs
```diff
@@ -0,0 +1,90 @@
...
+
+use crate::sync::{Arc, AtomicUsize};
+
+pub(crate) type Epoch = Arc<CachePadded<AtomicUsize>>;
+
+/// Passes copies of [`ReadHandle`]'s [`Epoch`]s to writer.
+///
+/// [`ReadHandle`]s don't clean up anything on [`drop`], so there is opportunity for an unbounded
+/// memory growth in case [`WriteHandle`] does nothing and new [`ReadHandle`]s are created and
+/// dropped.
```
ilslv, line 18: I would say this is my main concern with the implementation. Would this be a dealbreaker?

.github/workflows/check.yml
```diff
@@ -120,7 +120,7 @@ jobs:
     # https://docs.github.com/en/actions/learn-github-actions/contexts#context-availability
     strategy:
       matrix:
-        msrv: ["1.60.0"]
+        msrv: ["1.66.0"]
```
ilslv, line 123: `cordyceps` declares MSRV at `1.61.0`, but actually it's `1.66.0`. I'll report it to upstream a little later

src/epochs.rs
jonhoo: It's definitely a concern, but not necessarily a blocker.

jonhoo:
I'm hesitant to add something like this, because in general it's an anti-pattern to be doing a significant amount of creation and dropping of read handles. While it won't necessarily impact read performance, it'll be _terrible_ for the writer. And it suggests something in the application's design isn't quite right. So having the mutex there I think is in that sense a "feature". It also means we avoid the complexity of needing to make that part (look) concurrent.

PR #147, opened 2026-05-09
Make loom a dev-dependency

djc:
This avoids it leaking into downstream `Cargo.lock` files, for example via the recently added metrics-exporter-prometheus -> evmap -> left-right edge.


jonhoo (approved):
Good catch, thanks!

jonhoo:
Ah, actually, no, loom can't be a dev-dependency, because it also affects the types used in the library-under-test. Sorry!

PR #153, opened 2026-07-18
Mitigate race condition in WriteHandle::drop

jonhoo:
Also affects `WriteHandle::take`.

Previously, the `wait` in `take_inner` could return if some (new) read handles had only observed the last `publish`, but *not* the `swap` to `NULL`. Fix this by reading the epochs again after the `NULL`-swap and then wait to observe them tick over.

Closes #144.

Replaces #144 since I could not push to their PR branch, but wanted to make some changes to the comments


jonhoo:
Released in 0.11.8 :tada:

PR #154, opened 2026-07-18
Relax reentrant nullpointer assumption

jonhoo:
It's possible for `ReadHandle::enter` to observe `null` even after an initial successful `enter`, specifically when `WriteHandle::drop` is running concurrently, so the `unreachable!` was a little too strong.

Replaces #150 since I wanted to add a test and modify the comments, but the PR did not allow maintainer pushes.


jonhoo:
Released in 0.11.8 :tada:

